PR #27, opened 2012-11-13
Interactive Markers for setting objects pose

marioprats:
This pull request adds the functionality for setting an object pose (when creating a planning scene) using interactive markers. It is a new layer on top of the existing manual approach where the user could introduce the position and orientation numbers directly. Now the user can do it manually as before, or dragging the interactive marker, which automatically fills the position and orientation fields. The interactive markers are created by a direct instantiation of the rviz InteractiveMarker class, without requiring a server and a display. This is done thanks to the InteractiveMarker class refactoring recently introduced in rviz.

### DIFF
--- a/visualization/moveit_rviz_plugin/include/moveit_rviz_plugin/planning_frame.h
+++ b/visualization/moveit_rviz_plugin/include/moveit_rviz_plugin/planning_frame.h
@@ -38,6 +38,9 @@
 #include <move_group_interface/move_group.h>
 #include <moveit/warehouse/planning_scene_storage.h>
 #include <planning_scene_monitor/planning_scene_monitor.h>
+#include <interactive_markers/interactive_marker_server.h>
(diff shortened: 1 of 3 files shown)


isucan:
Awesome! Thanks for doing this Mario. There is a small problem with linking however. We now need to link against the default_plugins lib from rviz. That means that the default plugins lib gets loaded automatically when linking against the _core library of the moveit rviz plugin, not using the expected pluginlib dlopen. Is it possible to move the code used from the default_plugin library to the interactive markers package? It does not have to be done in this pull request, but I'd like to make sure that can be done before the groovy release, because porting these changes to groovy will cause some plugin issues if we do not fix this.

adamantivm:
@isucan is this going to be available in tag reachability2?

isucan:
in just a few minutes, yes.

PR #46, opened 2012-11-28
Updating location for denied location collada file

adamantivm:
Interacting with the end effector was failing due to an exception finding this collada file. This was making it almost impossible to define a goal or target pose.


adamantivm:
ahhh!! @isucan  is too fast! I'm submitting a new pull request to also account for install location as discovered by @marioprats before

PR #126, opened 2013-01-07
Support for differential IK

marioprats:


planning_interface/robot_interaction/include/moveit/robot_interaction/robot_interaction.h
```diff
@@ -85,6 +85,8 @@ class RobotInteraction
   class InteractionHandler
   {
   public:
+    typedef enum {POSITION_IK, VELOCITY_IK} InteractionModeType;
```
isucan: Do we really need interaction modes? Can we not just call the right function?

planning_interface/robot_interaction/include/moveit/robot_interaction/robot_interaction.h
```diff
@@ -150,14 +157,25 @@ class RobotInteraction
 
     bool transformFeedbackPose(const visualization_msgs::InteractiveMarkerFeedbackConstPtr &feedback, geometry_msgs::PoseStamped &tpose);
 
+    /** \brief Secondary task that tries to keep away from joint limits
+     * @param joint_state_group the joint state group for which to compute the task
+     * @param stvector the output of the function: a vector with joint velocities
+     * @param activation_threshold A percentage of the range from which the task is activated, i.e. activate if q > qmax - range * threshold. Typically between 0 and 0.5
+     * @param gain a gain for this task, multiplies the output velocities
+     */
+    bool avoidJointLimitsSecTask(const kinematic_state::JointStateGroup &joint_state_group, Eigen::VectorXd &stvector,
```
isucan: please pass JointStateGroup as pointer, just for consistency

planning_interface/robot_interaction/include/moveit/robot_interaction/robot_interaction.h
```diff
@@ -194,6 +212,7 @@ class RobotInteraction
   static bool updateState(kinematic_state::KinematicState &state, const EndEffector &eef, const geometry_msgs::Pose &pose,
                           unsigned int attempts, double ik_timeout, const kinematic_state::StateValidityCallbackFn &validity_callback = kinematic_state::StateValidityCallbackFn());
   static bool updateState(kinematic_state::KinematicState &state, const VirtualJoint &vj, const geometry_msgs::Pose &pose);
+  static bool updateState(kinematic_state::KinematicState &state, const EndEffector &eef, const geometry_msgs::Pose &pose, const kinematic_state::SecondaryTaskCallbackFn &st_callback);
```
isucan: Do we want to use geometry_msgs::Pose for this version? Should we not use the Twist message here?

planning_interface/robot_interaction/src/robot_interaction.cpp
```diff
@@ -75,12 +80,23 @@ void RobotInteraction::InteractionHandler::handleEndEffector(const robot_interac
...
-  
-  if (!robot_interaction::RobotInteraction::updateState(*kstate_, eef, tpose.pose, ik_attempts_, ik_timeout_, state_validity_callback_fn_))
+
+  bool update_state_result;
+  if (interaction_mode_ == POSITION_IK)
+  {
+    update_state_result = robot_interaction::RobotInteraction::updateState(*kstate_, eef, tpose.pose, ik_attempts_, ik_timeout_, state_validity_callback_fn_);
+  }
+  else if (interaction_mode_ == VELOCITY_IK)
+  {
```
isucan, line 151: I guess here is where we would convert to a Twist msg?

planning_interface/robot_interaction/src/robot_interaction.cpp
```diff
@@ -399,6 +415,120 @@ bool RobotInteraction::updateState(kinematic_state::KinematicState &state, const
   return state.getJointStateGroup(eef.parent_group)->setFromIK(pose, eef.parent_link, attempts, ik_timeout, validity_callback);
 }
 
+//TODO: move functions like this into an eigen_utils package
```
isucan: What does this function do? Do we really need an eigen_utils pkg? What else will we have there? If used only in this .cpp, please put this function in an anonymous namespace.

planning_interface/robot_interaction/src/robot_interaction.cpp
```diff
@@ -399,6 +415,120 @@ bool RobotInteraction::updateState(kinematic_state::KinematicState &state, const
...
+  for (std::size_t i = 0; i < qlimits.size(); ++i)
+  {
+    qmin(i) = qlimits[i].min_position;
+    qmax(i) = qlimits[i].max_position;
+    qrange(i) = qmax(i) - qmin(i);
+
+    //Fill in stvector with the gradient of a joint avoidance cost function
+    if (qrange(i) == 0)
+    {
+      //If joint range is zero, do not compute the cost. TODO: Is this always the case for continuous joints? should be set to zero too
```
isucan: Joint limits for continuous joints are not 0. You need to check the joint type. Continuous joints will report -pi to pi as joint limits, but that can be a prismatic joint as well.

visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp
```diff
@@ -664,7 +667,7 @@ void MotionPlanningDisplay::displayMetrics(bool start)
   }
 }
 
-void MotionPlanningDisplay::changedQueryStartState(void)
+void MotionPlanningDisplay::drawQueryStartState(void)
```
isucan, line 678: Are we sure the business you changed here with the update of the markers is OK? Please test with the PR2 and see what happens when attempting to set invalid IK poses.

planning_interface/robot_interaction/include/moveit/robot_interaction/robot_interaction.h
isucan: Ok, reading through the code i see that this makes sense.  I'd rename the type to IKInteractionType  because the InteractionHandler encompasses more than just stuff that is specific to IK.

planning_interface/robot_interaction/src/robot_interaction.cpp
marioprats: This converts an transform given in an homogeneous matrix to a pose vector in (t, utheta) representation. Eigen lacks a lot of functions useful for robotics. I would start with this one, the moore-penrose pseudoinverse, and twist transformation matrices, but for sure there are more that would be nice to have.

visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp
marioprats, line 678: Well, the arm just remains fixed at the last valid ik, but you can still move the marker around. When the marker goes back to the reachable space, the arm jumps to it.

planning_interface/robot_interaction/src/robot_interaction.cpp
marioprats: JointModel::JointType contains UNKNOWN, REVOLUTE, PRISMATIC, PLANAR, FLOATING, FIXED. Is a continuous joint of REVOLUTE type with limits -pi / pi ??

visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp
isucan, line 678: Can we have an option to make the marker jump to the last valid IK pose? It can be a 'reset'-like functionality. Maybe Ctrl + R or smth like that?

planning_interface/robot_interaction/src/robot_interaction.cpp
isucan: Would the eigen_conversions package be a suitable location for these
functions? It may be a little far fetched .. if it is, we will just start a
new package (outside moveit, so others can depend on it without moveit)

visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp
marioprats, line 678: done

PR #158, opened 2013-01-18
Adds threading to robot_interaction

aleeper:
Hopefully it doesn't suck :)

### DIFF
--- a/planning_interface/robot_interaction/include/moveit/robot_interaction/robot_interaction.h
+++ b/planning_interface/robot_interaction/include/moveit/robot_interaction/robot_interaction.h
@@ -170,6 +170,7 @@ public:
                                      double activation_threshold, double gain) const;
 
     std::string name_;
+    std::string planning_frame_;
     kinematic_state::KinematicStatePtr kstate_;
     boost::shared_ptr<tf::Transformer> tf_;
     std::set<std::string> error_state_;
@@ -226,7 +227,6 @@ private:
   void processInteractiveMarkerFeedback(const visualization_msgs::InteractiveMarkerFeedbackConstPtr& feedback);
   void processingThread(void);
 
-  // new stuff for processing thread
   boost::scoped_ptr<boost::thread> processing_thread_;
   bool run_processing_thread_;
 
@@ -234,7 +234,6 @@ private:
   boost::condition_variable new_action_condition_;
   std::map<std::string, visualization_msgs::InteractiveMarkerFeedbackConstPtr> feedback_map_;
 
-  // original stuff
   kinematic_model::KinematicModelConstPtr kmodel_;
   
   std::vector<EndEffector> active_eef_;
--- a/planning_interface/robot_interaction/include/moveit/robot_interaction/robot_interaction.h
+++ b/planning_interface/robot_interaction/include/moveit/robot_interaction/robot_interaction.h
@@ -35,6 +35,8 @@
 #include <visualization_msgs/InteractiveMarkerFeedback.h>
 #include <moveit/kinematic_state/kinematic_state.h>
 #include <boost/function.hpp>
+#include <boost/thread.hpp>
+#include <boost/thread/recursive_mutex.hpp>
 #include <tf/tf.h>
 
 namespace interactive_markers
@@ -103,13 +105,14 @@ public:
       return name_;
     }
     
-    const kinematic_state::KinematicStatePtr& getState(void) const
+    const kinematic_state::KinematicStatePtr& getState(void)
     {
       return kstate_;
     }
     
     void setState(const kinematic_state::KinematicState& kstate)
     {
+      boost::recursive_mutex::scoped_lock slock(state_lock_);
       *kstate_ = kstate;
     }    
     
@@ -176,6 +179,8 @@ public:
     double ik_timeout_;
     unsigned int ik_attempts_;
     IKInteractionType interaction_mode_;
+
+    boost::recursive_mutex state_lock_;
     
   private:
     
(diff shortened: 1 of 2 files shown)


planning_interface/robot_interaction/include/moveit/robot_interaction/robot_interaction.h
```diff
@@ -176,6 +179,8 @@ class RobotInteraction
     double ik_timeout_;
     unsigned int ik_attempts_;
     IKInteractionType interaction_mode_;
+
+    boost::recursive_mutex state_lock_;
```
isucan, line 184: Looking at the code, it looks like we do not need a recursive mutex. Can we switch to a simple mutex?

planning_interface/robot_interaction/include/moveit/robot_interaction/robot_interaction.h
```diff
@@ -35,6 +35,8 @@
 #include <visualization_msgs/InteractiveMarkerFeedback.h>
 #include <moveit/kinematic_state/kinematic_state.h>
 #include <boost/function.hpp>
+#include <boost/thread.hpp>
+#include <boost/thread/recursive_mutex.hpp>
```
isucan, line 39: ... and change the include

planning_interface/robot_interaction/include/moveit/robot_interaction/robot_interaction.h
```diff
@@ -218,8 +223,18 @@ class RobotInteraction
...
+
+  // new stuff for processing thread
+  boost::scoped_ptr<boost::thread> processing_thread_;
+  bool run_processing_thread_;
+
+  mutable boost::mutex action_lock_;
+  boost::condition_variable new_action_condition_;
+  std::map<std::string, visualization_msgs::InteractiveMarkerFeedbackConstPtr> feedback_map_;
+
+  // original stuff
```
isucan: please remove this comment

planning_interface/robot_interaction/src/robot_interaction.cpp
```diff
@@ -135,13 +135,19 @@ void eigenTransformToEigenVector(const Eigen::Affine3d &M, Eigen::VectorXd &pose
 void RobotInteraction::InteractionHandler::handleEndEffector(const robot_interaction::RobotInteraction::EndEffector& eef,
                                                              const visualization_msgs::InteractiveMarkerFeedbackConstPtr &feedback)
 { 
-  if (feedback->event_type != visualization_msgs::InteractiveMarkerFeedback::POSE_UPDATE)
-    error_state_.clear();
+  // aleeper: I think this may have been left over from when you wanted the arms to automatically jump back?
+  //          I see no reason this should be here, and it is causing problems with my attempts
+  //          to update the displayed robot at the correct times.
+  // if (feedback->event_type != visualization_msgs::InteractiveMarkerFeedback::POSE_UPDATE)
+  //   error_state_.clear();
```
isucan: true. please remove the code + comment

planning_interface/robot_interaction/src/robot_interaction.cpp
```diff
@@ -198,7 +206,9 @@ bool RobotInteraction::InteractionHandler::inError(const robot_interaction::Robo
 
 bool RobotInteraction::InteractionHandler::transformFeedbackPose(const visualization_msgs::InteractiveMarkerFeedbackConstPtr &feedback, geometry_msgs::PoseStamped &tpose)
 {
+  state_lock_.lock();
```
isucan: I think we do not need to lock here, since we just need a value that does not change. We should initialize a planning_frame_ variable in the constructor of this class, and use it here. No need to lock in the constructor.

planning_interface/robot_interaction/src/robot_interaction.cpp
```diff
@@ -556,22 +575,74 @@ void RobotInteraction::processInteractiveMarkerFeedback(const visualization_msgs
...
+void RobotInteraction::processingThread(void)
+{
+  boost::unique_lock<boost::mutex> ulock(action_lock_);
+
+  while (run_processing_thread_ && ros::ok())
   {
-    ROS_ERROR("Interactive Marker Handler '%s' is not known.", handler_name.c_str());
-    return;
+    // This timed_wait is needed so that the node goes down quickly when requested.
+    boost::system_time const timeout=boost::get_system_time()+ boost::posix_time::milliseconds(100);
```
isucan: please use consistent spacing around +
and use spacing around =

planning_interface/robot_interaction/src/robot_interaction.cpp
```diff
@@ -556,22 +575,74 @@ void RobotInteraction::processInteractiveMarkerFeedback(const visualization_msgs
...
+    while (feedback_map_.empty() && run_processing_thread_ && ros::ok())
+    {
+      new_action_condition_.timed_wait(ulock, timeout);
```
isucan: does this really need to be timed wait? is there a case when the condition never gets broadcasted?

isucan:
Hey Adam,

This looks great actually!
I only added some minor comments to the code.

Ioan

planning_interface/robot_interaction/include/moveit/robot_interaction/robot_interaction.h
aleeper, line 184: It needs to be recursive because the setState() function also locks. Since handleEndEffector calls a user-supplied callback, wherein the user can then call setState(), it can get stuck if the mutex isn't recursive.

isucan, line 184: ok, sounds good.

aleeper:
Ok the timed wait wasn't needed after all. I thought it was needed to make sure the program quits quickly, but I just tested it without that.

PR #170, opened 2013-01-23
Make moveit_warehouse_save_as_text save the queries as well as the scenes

ibrahima:
This allows moveit_warehouse_save_as_text to save the initial robot states and queries in addition to saving the scenes.

### DIFF
--- a/warehouse/warehouse/src/save_as_text.cpp
+++ b/warehouse/warehouse/src/save_as_text.cpp
@@ -48,25 +48,25 @@ static const std::string ROBOT_DESCRIPTION="robot_description";
 typedef std::pair<geometry_msgs::Point, geometry_msgs::Quaternion> LinkConstraintPair;
 typedef std::map<std::string, LinkConstraintPair > LinkConstraintMap;
 
-LinkConstraintMap collectLinkConstraints(moveit_msgs::Constraints constraints)
+void collectLinkConstraints(const moveit_msgs::Constraints& constraints, LinkConstraintMap& lcmap )
 {
-  LinkConstraintMap lcmap;
-  for(int i=0; i<constraints.position_constraints.size(); i++){
+  for(std::size_t i = 0; i < constraints.position_constraints.size(); ++i)
+  {
     LinkConstraintPair lcp;
     const moveit_msgs::PositionConstraint &pc = constraints.position_constraints[i];
     lcp.first = pc.constraint_region.primitive_poses[0].position;
     lcmap[constraints.position_constraints[i].link_name] = lcp;
   }
 
-  for(int i=0; i<constraints.orientation_constraints.size(); i++){
-    if(lcmap.count(constraints.orientation_constraints[i].link_name)){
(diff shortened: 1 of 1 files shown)


ibrahima:
I hope this fits in with ROS coding standards and whatnot. If not, I'd be happy to fix it.

One thing to note is that currently this will generate a queries file for all scenes in the warehouse, even if they don't have any associated queries. It also currently outputs all joints in the RobotState just because that's what KinematicState::printStateInfo does, although it might be nice to remove joints that are at 0. It seems to generate the same output as the input files that I imported, except for not printing as many digits of the joint states.

isucan:
This is fantastic! Thanks! I will put a few comments wrt style if you don't mind, but the code is great! Thanks!
There is no need to remove joints that are at 0, but it would be nice not to generate empty files.

warehouse/warehouse/src/save_as_text.cpp
```diff
@@ -35,14 +35,40 @@
...
-
+#include <moveit/kinematic_state/conversions.h>
 #include <ros/ros.h>
 
 static const std::string ROBOT_DESCRIPTION="robot_description";
 
+typedef std::pair<geometry_msgs::Point, geometry_msgs::Quaternion> LinkConstraintPair;
+typedef std::map<std::string, LinkConstraintPair > LinkConstraintMap;
+
+LinkConstraintMap collectLinkConstraints(moveit_msgs::Constraints constraints)
```
isucan: It is more efficient to pass the argument as const moveit_msgs::Constraints& constraints;
Also, it would be better if LinkConstraintMap was passed as an argument (LinkConstraintMap&), to prevent a copy on the stack. (Not sure if all compilers use move constructors yet)

warehouse/warehouse/src/save_as_text.cpp
```diff
@@ -35,14 +35,40 @@
...
+{
+  LinkConstraintMap lcmap;
+  for(int i=0; i<constraints.position_constraints.size(); i++){
```
isucan: To prevent warnings it is nice to have the index variable be the same type as that returned by size(); This is std::size_t 
I would write it as:
for (std::size_t i = 0 ; i < constraints.position_constraints.size() ; ++i)

(I also prefer spaces around operators and prefix ++)

warehouse/warehouse/src/save_as_text.cpp
```diff
@@ -35,14 +35,40 @@
...
+    LinkConstraintPair lcp;
+    const moveit_msgs::PositionConstraint &pc = constraints.position_constraints[i];
+    lcp.first = pc.constraint_region.primitive_poses[0].position;
+    lcmap[constraints.position_constraints[i].link_name] = lcp;
+  }
+
+  for(int i=0; i<constraints.orientation_constraints.size(); i++){
```
isucan: Same issue as above for the for loop

ibrahima:
OK, I think I've addressed everything.

isucan:
Awesome. Thanks!

PR #173, opened 2013-01-24
Set the link name on constraints when saving an end effector goal

ibrahima:
Right now it seems that in the RViz plugin when you save an end effector goal to the database it doesn't actually set the link_name field of the constraints. This fixes that.

### DIFF
--- a/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_queries.cpp
+++ b/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_queries.cpp
@@ -115,6 +115,7 @@ void MotionPlanningFrame::createGoalPoseButtonClicked(void)
           posemsg.orientation.w = 1.0;
           pc.constraint_region.primitive_poses.push_back(posemsg);
           pc.weight = 1.0;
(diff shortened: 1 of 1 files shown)


ibrahima:
Actually, is this intentional? I guess the benchmark config specifies a default constrained link so it's not really necessary to have link_name specified by the constraint for benchmarking purposes. However this does cause the import script to fail when run on the output generated by my query saving code because it skips over the blank line when parsing the file. It seems like regardless of whether a link_name is specified in the constraint I can still give it as a goal to either arm so I'm guessing it's fine to do this.

isucan:
Actually, the link name not being specified is indeed intentional. This is so the poses can be re-used for both pr2 arms, for example, or for different robot arms. The empty line issue can be resolved i guess with a call to getline()? and checking if the line is empty?

ibrahima:
Ah I see. Maybe peek()ing at the next character instead? Just because otherwise you'd have to specify a maximum length for that field and handle what happens if the line is longer than what you specified.

isucan:
I like the way you think :) That sounds good.

PR #175, opened 2013-01-24
Handle blank link names in imported queries files.

ibrahima:


ibrahima:
Wait, does this actually work? I think I broke importing constraints when the link name is specified, actually, because peek() isn't really what you want because there's always a new line. I can throw together a fix sometime soonish, although since it's not actually preventing the queries from being imported (just gives an error about the link name not being a valid link constraint element) I'll probably do this later.

isucan:
Good point. You should use getline(). Use this variant though:
http://www.cplusplus.com/reference/string/string/getline/
Which avoids the buffer length issue.

Issue #177, opened 2013-01-25
render_shapes holds only one ogre material for all meshes


marioprats:
...and this ogre material is initialized with the color of the first drawn mesh. As a result, all the meshes are rendered with the same color, even if their colors are actually different. This just happens for meshes. The rest of shapes primitives are rendered correctly.

I guess we should keep a vector of materials, and assign one to each new mesh. If this makes sense I can send a pull request.

isucan:
That makes sense. Pull request welcome :)

PR #178, opened 2013-01-25
render shapes fix to create all the materials for meshes

marioprats:


marioprats:
fixes #177

visualization/planning_scene_rviz_plugin/src/render_shapes.cpp
```diff
@@ -109,32 +110,30 @@ void RenderShapes::renderShape(Ogre::SceneNode *node, const shapes::Shape *s, co
       const shapes::Mesh *mesh = static_cast<const shapes::Mesh*>(s);
       if (mesh->triangle_count > 0)
       {
-        // check if we need to construct the material
-        if (material_name_.empty())
+        //construct the material
+        std::string mname = "Planning Scene Display Mesh Material @" + boost::lexical_cast<std::string>(materials_.size());
```
isucan: This is not correct. If you have a PlanningScene and a MotionPlanning plugin in the same rviz process the line above will produce 2 materials with the same name. Please re-add the use of "this" in the construction of the name, in addition to the material count.

visualization/planning_scene_rviz_plugin/src/render_shapes.cpp
```diff
@@ -109,32 +110,30 @@ void RenderShapes::renderShape(Ogre::SceneNode *node, const shapes::Shape *s, co
...
+          mat->getTechnique(0)->setSceneBlending( Ogre::SBT_REPLACE );
+          mat->getTechnique(0)->setDepthWriteEnabled( true );
         }
+        materials_.push_back(mat);
 
         std::string name = "Planning Scene Display Mesh " + boost::lexical_cast<std::string>(movable_objects_.size()) + " @" + boost::lexical_cast<std::string>(this);
         Ogre::ManualObject *manual_object = context_->getSceneManager()->createManualObject(name);
         manual_object->estimateVertexCount(mesh->triangle_count * 3);
-        manual_object->begin(material_name_, Ogre::RenderOperation::OT_TRIANGLE_LIST);
+        manual_object->begin(materials_[materials_.size() - 1]->getName(), Ogre::RenderOperation::OT_TRIANGLE_LIST);
```
isucan: It is a little more elegant (and slightly more efficient) to say materials_.back()->getName() instead of materials_[materials_.size() - 1]->getName()

visualization/planning_scene_rviz_plugin/src/render_shapes.cpp
marioprats: Right! I didn't realize we can have two different plugins using render_shapes, and was wondering what was that "this" for!

PR #184, opened 2013-01-30
First version of benchmark tool

marioprats:
This is a first functional version of the benchmark GUI. For now it just includes the functionality we had in the moveit rviz plugin, plus the capability of switching between robots without leaving the app (File->Load Robot). I'll close the related issue and start opening more concrete tickets for bugs and features.


marioprats:
Please note that this will require all to update to the latest debs (because of pluginlib compatibility with boost filesystem v2)

PR #195, opened 2013-02-07
States tab in moveit rviz plugin

marioprats:
This implements functionality in the states tab of the moveit rviz plugin

### DIFF
--- a/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_states.cpp
+++ b/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_states.cpp
@@ -1,0 +1,214 @@
+/*
+ * Copyright (c) 2013, Willow Garage, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *     * Neither the name of the Willow Garage, Inc. nor the names of its
+ *       contributors may be used to endorse or promote products derived from
+ *       this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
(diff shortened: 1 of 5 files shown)


visualization/motion_planning_rviz_plugin/src/motion_planning_frame_states.cpp
```diff
@@ -0,0 +1,236 @@
...
+  saveRobotStateButtonClicked(*planning_display_->getQueryGoalState());
+}
+
+void MotionPlanningFrame::setAsStartStateButtonClicked()
+{
+  QListWidgetItem *item = ui_->list_states->currentItem();
+
+  if (item)
+  {
+    robot_state::RobotState robot_state(planning_display_->getKinematicModel());
```
isucan: In case the msg you read from the db is incomplete (no joints specified), we should use the state we are about to replace as initialization.
so use 
robot_state::RobotState robot_state(*planning_display_->getQueryStartState());

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_states.cpp
```diff
@@ -0,0 +1,236 @@
...
+  }
+}
+
+void MotionPlanningFrame::setAsGoalStateButtonClicked()
+{
+  QListWidgetItem *item = ui_->list_states->currentItem();
+
+  if (item)
+  {
+    robot_state::RobotState robot_state(planning_display_->getKinematicModel());
```
isucan: use robot_state::RobotState robot_state(*planning_display_->getQueryGoalState());

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_states.cpp
```diff
@@ -0,0 +1,236 @@
...
+    msgBox.setStandardButtons(QMessageBox::Yes | QMessageBox::No | QMessageBox::Cancel);
+    msgBox.setDefaultButton(QMessageBox::No);
+    int ret = msgBox.exec();
+
+    switch (ret)
+    {
+      case QMessageBox::Yes:
+      {
+        QList<QListWidgetItem*> found_items =  ui_->list_states->selectedItems();
+        for (unsigned int i = 0; i < found_items.size() ; ++i)
```
isucan: use std::size_t instead of unsigned int

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_states.cpp
```diff
@@ -0,0 +1,236 @@
...
+      {
+        robot_state_storage_->getKnownRobotStates(text.toStdString(), names);
+      }
+      catch (std::runtime_error &ex)
+      {
+        QMessageBox::warning(this, "Cannot query the database", QString("Wrongly formatted regular expression for robot states: ").append(ex.what()));
+        return;
+      }
+
+      for ( unsigned int i = 0 ; i < names.size() ; i++ )
```
isucan: use std::size_t instead of unsigned int;
in general prefix (++i) is desirable over (i++) although the compiler probably optimizes that out

PR #213, opened 2013-03-19
Nearly trivial patch to planning_scene_interface

mikeferguson:


isucan:
This api is not too great right now and i was planning on changing it. What
are you using it for?

mikeferguson:
I'm not actually using it right now -- I was just using it for testing, and noticed that removing objects failed to work, and tracked down this change.

isucan:
Alright. I expect this will be fixed in the next few weeks.

Ioan

PR #242, opened 2013-05-13
Changes from rob@work

ipa-raw3-1:
This is a wrap_up/code_commit after rob@work's ICRA/ROSCON demo playing back the changes to ros-planning


warehouse/warehouse/src/initialize_demo_db.cpp
```diff
@@ -122,14 +122,14 @@ int main(int argc, char **argv)
         
         moveit_msgs::OrientationConstraint ocm;
         ocm.link_name = lnames.back();
-        ocm.header.frame_id = psm.getRobotModel()->getModelFrame();
```
isucan, line 125: can we exclude this change?

planning_interface/move_group_interface/include/moveit/move_group_interface/move_group.h
```diff
@@ -326,6 +326,33 @@ class MoveGroup
   /**@}*/
 
   /**
+   * \defgroup follow_existing_traj Follow an existing sequence of constraints
```
isucan, line 329: It looks like you have forked from an older version and somehow git wants to re-add stuff I have removed; could you rebase?

Issue #140, opened 2013-01-10
Ability to add menu callbacks to the interactive markers in robot_interaction.


aleeper:
Menus are very useful. Robot interaction should have an interface for defining menus to add to a marker along with an associated callback.

aleeper:
Just adding this issue to the alpha milestone. I think this is an important feature if we mean to let people write flexible applications using MoveIt.

aleeper:
Closing as #247 addressed this.

PR #247, opened 2013-05-21
Adds ability to set a menu handler in robot interaction

aleeper:
After much hemming and hawing, I decided the cleanest way to do menus right now is to simply allow the user to set a single menu handler for each interaction handler. Hence, different subgroups in the currently active group will all get the same set of menus. When the user changes the active group, they could change the menu handler to have different menu entries if desired. I think this is a good balance between flexibility and simplicity of API.

To show how it works, there is basic functionality added to the motion planning plugin. It creates a menu handler for the start state, and a separate one for the goal state. Now, you can right click on the markers to change planning groups, or to set the state that you have right-clicked-on to random, current, or the "other" one (start or goal). 

Try it out and let me know what you think.
#140


isucan:
This looks good. I'll make a small adjustment after merge, but things should not differ too much

PR #255, opened 2013-06-05
adding removeSlash here as well to deal with frames with a leading slash

sachinchitta:


isucan:
The way we deal with frames now should worfine with slash. In fact, the planning frame should have a slash. Can you describe the problem you found? I think the fix needs to be different.

sachinchitta:
This comes up when I am using the benchmarks gui - robot_interaction looks for the link state for "/odom_combined" and keeps printing an error (line 341 of robot_interaction.cpp). It is not supposed to look for the link state for that frame since it is the planning frame. However, because of the presence of the extra slash, the comparison test for this fails.

isucan:
That is a bug in benchmark_gui then. Frame names are not necessarily states; I'll take a look at the benchmark_gui to fix this; If you have a location in the code, that will help as well. The static function Transforms::isSameFrame() is what we are supposed to use now, to compare frame names.

PR #266, opened 2013-06-18
added missing getRobotRootLink python wrapper

mathias-luedtke:
needed for ground command in moveit_commander


isucan:
Mathias, the code you added used to be in the bindings, but then got moved to the robot_commander class. This is because the function -- getting a robot's root link -- is not specific to a particular group. Is it possible to call the function from the new location instead?

mathias-luedtke:
Ah, I see..
Then it was a bug in move_commander, see https://github.com/ros-planning/moveit_commander/pull/5

The function declaration for getRobotRootLink should be removed from the header file..
And what's about getPlanningFrame?

isucan:
Good point about getPlanningFrame. Fixing now.

PR #269, opened 2013-06-27
Make KDL Plugin searchPositionIK actually const, and thread-safe

mikeferguson:
This fixes the KDL plugin to close https://github.com/ros-planning/moveit_core/issues/47

I've tested the IK on both my robot Maxwell, as well as the PR2 arm+torso (which uses the KDL plugin) and all looks good. I've tested that I can send pick/place requests with many grasps/place-locations without any issue on Maxwell.

The added steps to instantiate the JntArrays and IK/FK solvers in each call to searchPositionIK do not appear slow things down noticeably.

### DIFF
--- a/planning/kdl_kinematics_plugin/include/moveit/kdl_kinematics_plugin/kdl_kinematics_plugin.h
+++ b/planning/kdl_kinematics_plugin/include/moveit/kdl_kinematics_plugin/kdl_kinematics_plugin.h
@@ -67,7 +67,7 @@
 #include <moveit/robot_model/robot_model.h>
 #include <moveit/robot_state/robot_state.h>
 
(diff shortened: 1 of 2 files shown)


isucan:
Thanks Mike! Very nice of you to dig and fix this!

sachinchitta:
One of the other requirements on the kinematics solver, apart from needing it to be thread-safe, is also realtime safety. I am pretty sure the old implementation was realtime safe (except the constructor) but the new one is no longer so since its allocating stuff with every call. We will have to re-examine the implementation of this to make sure its realtime safe.

mikeferguson:
@sachinchitta Why is realtime safe a requirement for this component?

PR #272, opened 2013-07-03
Added options struct to kinematics, KDL kinematics

sachinchitta:


mikeferguson:
@isucan An email about this should definitely go out to the mailing list if/when it is merged -- as it will break every IKfast plugin that has been generated to date. @davetcoleman note my previous statement.

isucan:
@mikeferguson an e-mail will be sent for sure

isucan:
https://groups.google.com/forum/#!topic/moveit-users/948Z-CD8WSw

PR #288, opened 2013-08-06
benchmarks gui: select planners, nb of runs and timeout

marioprats:
This will add the possibility to select the planners to benchmark, number of runs and timeout from the benchmarks GUI.


isucan:
thanks!

Issue #304, opened 2013-09-07
Why absolute default topic names


bit-pirate:
My MoveIt stopped working, when I started to use the newest source.
Looking further into it, I realised that some topics are not connected any more.

The related topics are the ones from the planning scene monitor, which are subscribed to using this root handle, and from the current state monitor, i.e. joint states.

What did exactly change since the released version and why?

For our use cases we usually put all robot related topics in the robot name space, e.g. using the roslaunch's group tag. Unfortunately that doesn't work anymore.

isucan:
Nothing changed recently with topics. Can you be more specific about which topics are not connected?

bit-pirate:
Sorry, just figured out what really went wrong: Looks like the new version uses _absolute_ default topic names. Is there a reason for that?

Removing the trailing "/" solves this issue for me.

PS: I'm still using a several weeks old moveit config. Maybe the new setup assistant solves this issue.

isucan:
Can you please be more specific about which topics you are talking about?

bit-pirate:
Sure, I adjusted the default topics below in `moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp`.

```
const std::string planning_scene_monitor::PlanningSceneMonitor::DEFAULT_JOINT_STATES_TOPIC = "joint_states";
const std::string planning_scene_monitor::PlanningSceneMonitor::DEFAULT_ATTACHED_COLLISION_OBJECT_TOPIC = "attached_collision_object";
const std::string planning_scene_monitor::PlanningSceneMonitor::DEFAULT_COLLISION_OBJECT_TOPIC = "collision_object";
const std::string planning_scene_monitor::PlanningSceneMonitor::DEFAULT_PLANNING_SCENE_WORLD_TOPIC = "planning_scene_world";
const std::string planning_scene_monitor::PlanningSceneMonitor::DEFAULT_PLANNING_SCENE_TOPIC = "planning_scene";
const std::string planning_scene_monitor::PlanningSceneMonitor::MONITORED_PLANNING_SCENE_TOPIC = "monitored_planning_scene";
```

PS: I'm sorry, I got confused. This change was already in the released (Hydro) version. I just made the jump from Groovy to Hydro and now going back and forward between the released and hydro-devel versions.

PR #305, opened 2013-09-07
planning_scene_monitor: adds log ouput makeup

bit-pirate:
Small addition for better topic name log output related to #304


isucan:
Thanks!

PR #310, opened 2013-09-12
specifies python version 2.7 for linking (fixes #302)

### LINKED ISSUE #302: Broken libpython3.2-dev deb crashes source built
For my hydro source install on Precise I had to manually install python3.2-dev to get the planning interface compiling.

Shouldn't that be listed in the package.xml/CMakeLists.txt?

bit-pirate:


isucan:
should the choice of 2.7 always be used? Should this be done on a per-distro basis?

bit-pirate:
Yes, in general a distro-based version would be better. However, since 2.7 will be the default until 14.04, there is now need for that for a while.

PS: I don't know how to do the checking for a distro, hence I didn't add it.

PR #322, opened 2013-09-30
fix segfault in approach translate

mikeferguson:
Not sure why we're iterating through the points and not just taking the last one, but we certainly can't iterate through joint names and then use the index with points.


manipulation/pick_place/src/approach_and_translate_stage.cpp
```diff
@@ -115,7 +115,7 @@ bool isStateCollisionFree(const planning_scene::PlanningScene *planning_scene,
   if (grasp_posture->joint_names.size() > 0)
   {
     // apply the grasp posture for the end effector (we always apply it here since it could be the case the sampler changes this posture)
-    for (std::size_t i = 0 ; i < grasp_posture->joint_names.size() ; ++i)
+    for (std::size_t i = 0 ; i < grasp_posture->points.size() ; ++i)
```
isucan, line 118: good catch!

PR #327, opened 2013-10-02
fix #324

### LINKED ISSUE #324: Unable to construct sampler when kinematics not loaded into ompl pipeline
The setup assistant no longer generates a line like the following in ompl_planning_pipeline.launch.xml:

```
<rosparam command="load" file="$(find robot_moveit_config)/config/kinematics.yaml"/>
```

This then causes an error when trying to use the pick interface:

```
[ INFO] [1380506826.859073575]: Planning attempt 1 of at most 1
[ INFO] [1380506826.859455926]: Added plan for pipeline 'pick'. Queue is now of size 1
[ERROR] [1380506826.870277203]: No sampler was constructed
[ INFO] [1380506826.870473586]: Manipulation plan 0 failed at stage 'reachable & valid pose filter' on thread 1
[ WARN] [1380506826.870695152]: All supplied grasps failed. Retrying last grasp in verbose mode.
[ INFO] [1380506826.870788668]: Re-added last failed plan for pipeline 'pick'. Queue is now of size 1
[ INFO] [1380506826.879020452]: Manipulation plan 0 failed at stage 'reachable & valid pose filter' on thread 0
[ INFO] [1380506826.879411293]: Pickup planning completed after 0.011495 seconds
```

I presume the plan was to load kinematics from robot_description_kinematics, and that something is missing, but I have no idea where that something is.

mikeferguson:


planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -227,7 +227,7 @@ void kinematics_plugin_loader::KinematicsPluginLoader::status() const
         ROS_DEBUG("Loading settings for kinematics solvers from the ROS param server ...");
 
         // read data using ROS params
-        ros::NodeHandle nh("~");
```
isucan, line 230: I can see why this change works with the default location for settings, but I am not sure it does in the local namespace of a node. For instance, if my params are under /foo/group/solver then doing a searchparam with group/solver in the root namespace won't find foo. But if we start that in /foo, it will. Maybe I do not understand params correctly, but when we're looking for robot_description_kinematics/group/solver in /foo, search param would test for /foo/robot_description_kinematics/group/solver, which is not there, but then test for /robot_description_kinematics/group/solver, which is. Does this make sense?

mikeferguson:
@isucan, good point-- I missed that case

The problem that I am running into is that move_group/arm exists, because we have planner info loaded, so search returns move_group/arm/kinematics_solver as the name and says it found it as true, so we load an empty param! I've changed this to be a check that we have the param returned by search. I've tested this against having move_group/arm/kinematics_solver and not having it, and we get the right results.

PR #340, opened 2013-10-12
fixes for mimic joints

sachinchitta:
This fixes the kdl kinematics solver for mimic joints. It also adds a kinematics query options map to robot_interaction allowing different query options to be set for different end-effectors.


planning/kdl_kinematics_plugin/src/kdl_kinematics_plugin.cpp
```diff
@@ -163,7 +163,7 @@ bool KDLKinematicsPlugin::initialize(const std::string &robot_description,
     return false;
   }
 
-  dimension_ = joint_model_group->getVariableCount();
+  dimension_ = joint_model_group->getJointModels().size();
```
isucan: This includes fixed joints too. I am not sure this is correct. You can get only the active joint models using getActiveJointModels(), which excludes mimic & fixed joints. You can get mimic joints with getMimicJointModels(). So you can use getActiveJointModels().size() + getMimicJointModels().size() to avoid including fixed joints.

robot_interaction/include/moveit/robot_interaction/robot_interaction.h
```diff
@@ -248,6 +248,21 @@ class RobotInteraction
...
+    void setKinematicsQueryOptionsForGroup(const std::string& group_name, 
+					   kinematics::KinematicsQueryOptions &options)
+    {
+      kinematics_query_options_map_[group_name] = options;
+    }
+
+    const bool getKinematicsQueryOptionsForGroup(const std::string& group_name, 
+						 kinematics::KinematicsQueryOptions &opt)
+    {
+      if(kinematics_query_options_map_.find(group_name) == kinematics_query_options_map_.end())
```
isucan: To avoid two lookups, we can use this instead:

```
std::map<std::string, kinematics::KinematicsQueryOptions>::const_iterator it = kinematics_query_options_map_.find(group_name);
if (it == kinematics_query_options_map_.end())
  return false;
opt = it->second;
return true;
```

sachinchitta:
Ok, I updated it.

PR #344, opened 2013-10-16
Manipulator_state_publisher and cartesian_teleoperation

ghost:
- manipulator_state_publisher: it takes a robot state and outputs a ros message consisting in the Pose and the Twist of the end_effector. The Twist is computed using subsequent joint values.
- cartesian_teleoperation: it takes desired Pose and Twist, performs IK, and publishes it to the joint_path_command topic.


isucan:
Thank you for the contribution. It seems to me some files are added multiple times (e.g., the .h file) in different locations. Can we fix that? Additionally, I'd like these to be libraries within the manipulation package, not separate packages on their own. Could we update the build system? Also, can we squash all the commits into one ?

mikeferguson:
@isucan should there also be license headers on the files? I know it's not entirely needed, but seeing as almost every other cpp/header has the license header perhaps it should?

isucan:
Oh yes, @mikeferguson is right. Thanks! All files should have the same header style: license + author name.

mikeferguson:
@isucan Also, the LICENSE file is a GNU GPL one...

isucan:
I believe the license is a typo -- Flavian and I already discussed the license situation (and all merged code will be BSD)

sachinchitta:
This is a great contribution, we have been thinking about adding a tele-operation interface to MoveIt! for a while. I do think we need to change a few choices before merging this.

(1) I keep seeing lgsm_conversions. Do we really need this? It seems like most of the work can be done without needing that package unless there's something there that I am missing. 
(2) I feel like this capability could natively go as a move_group capability instead of being a different package.
(3) I am not sure about this publishing as a joint path. Its only sending out one point at a time. There could be jumps in the ik solution (I don't see any interpolation going on). That will lead to weird behaviors on the robot - has this been tested with a particular robot?

ghost:
@isucan Please feel free to update the build system. As a beginner in catkin I tried my best but I would be interested in seeing how to make it cleaner. You can also squash the commit (I wanted to do this in the first place anyway).

@mikeferguson any licensing differing from BSD is a typo. As long as my name and the lab's (CEA LRI) appear, I am fine.

@sachinchitta Well I'm glad you like it! 
(1) We don't need this, I thought I had eliminated any reference to this package, apparently I missed a few.
(3-1) So we had a discussion with the swri group on that (https://groups.google.com/forum/#!topic/swri-ros-pkg-dev/rRL9vwQ8yK0, see 4) in Jeremy's first reponse to my first message). We figured this was the only easy way to adapt the driver to single-point trajectories.
(3-2) From what I gathered the IK solver tries to find the closest solution to the previous one. If your send rate is high enough, you should be fine. That holds only provided you are "far" from singularities of course. In my application I force the robot to go to a certain position where I know it has room to move so that was not my main concern. I guess you could check if the IK solution is not too far, and if it is, then maybe call a group.move() to put it in the right place (I don't think that's the best way to handle it though). This is an open question.
(3-3) About filtering, I think it would be interesting to add a "stream_filter" functionality or something like that.
(3-4) I have tested this on a sia10d with the motoman package from swri, using a slightly modified version that I pull requested. It ran at 50Hz.

sachinchitta:
We are not maintaining MoveIt! hydro any more.  I think this capability belongs better inside ros-controllers though.

PR #363, opened 2013-11-05
fixed dirty transform bugs

hersh:
This fixes 3 places where the benchmarks_gui crashes because of dirty transforms if you build in Debug mode.  If you build in Release mode, it doesn't crash but the markers show up at 0,0,0 instead of the end-effector.

### DIFF
--- a/benchmarks_gui/src/tab_states_and_goals.cpp
+++ b/benchmarks_gui/src/tab_states_and_goals.cpp
@@ -778,6 +778,7 @@ void MainWindow::copySelectedGoalPoses(void)
     std::stringstream ss;
     ss << scene_name.c_str() << "_pose_" << std::setfill('0') << std::setw(4) << goal_poses_.size();
 
+    scene_display_->getPlanningSceneRW()->getCurrentStateNonConst().update();
     Eigen::Affine3d tip_pose = scene_display_->getPlanningSceneRO()->getCurrentState().getGlobalLinkTransform(robot_interaction_->getActiveEndEffectors()[0].parent_link);
     geometry_msgs::Pose marker_pose;
     marker_pose.position.x = goal_poses_[name]->imarker->getPosition().x;
--- a/benchmarks_gui/src/tab_trajectories.cpp
+++ b/benchmarks_gui/src/tab_trajectories.cpp
@@ -82,6 +82,7 @@ void MainWindow::createTrajectoryButtonClicked(void)
       else
       {
         //Create the new trajectory starting point at the current eef pose, and attach an interactive marker to it
+        scene_display_->getPlanningSceneRW()->getCurrentStateNonConst().update();
         Eigen::Affine3d tip_pose = scene_display_->getPlanningSceneRO()->getCurrentState().getGlobalLinkTransform(robot_interaction_->getActiveEndEffectors()[0].parent_link);
         geometry_msgs::Pose marker_pose;
         tf::poseEigenToMsg(tip_pose, marker_pose);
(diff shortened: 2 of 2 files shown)


benchmarks_gui/src/tab_states_and_goals.cpp
```diff
@@ -778,6 +778,7 @@ void MainWindow::copySelectedGoalPoses(void)
     std::stringstream ss;
     ss << scene_name.c_str() << "_pose_" << std::setfill('0') << std::setw(4) << goal_poses_.size();
 
+    scene_display_->getPlanningSceneRW()->getCurrentStateNonConst().update();
```
isucan, line 781: This is actually not correct. You update the current state, and then use the current state, but changes could have come in in between. One way to do this would be something like this:

```
moveit::core::RobotState rs = scene_display_->getPlanningSceneRW()->getCurrentStateNonConst();
Eigen::Affine3d tip_pose = rs..getGlobalLinkTransform(robot_interaction_->getActiveEndEffectors()[0].parent_link);
// since rs is not const here, the getGlobalLinkTransform() call will trigger update() when needed
```

benchmarks_gui/src/tab_trajectories.cpp
```diff
@@ -82,6 +82,7 @@ void MainWindow::createTrajectoryButtonClicked(void)
       else
       {
         //Create the new trajectory starting point at the current eef pose, and attach an interactive marker to it
+        scene_display_->getPlanningSceneRW()->getCurrentStateNonConst().update();
```
isucan, line 85: same comment as above.

benchmarks_gui/src/tab_trajectories.cpp
```diff
@@ -337,6 +338,7 @@ void MainWindow::trajectoryExecuteButtonClicked()
     if (waypoint_poses.size() > 0)
     {
       robot_state::RobotState &rstate = scene_display_->getPlanningSceneRW()->getCurrentStateNonConst();
+      rstate.update();
```
isucan, line 341: we should just remove the `&` from the line above and the call to update() on this line.

benchmarks_gui/src/tab_states_and_goals.cpp
hersh, line 781: Hi Ioan, thanks for the correction.

Wow, well we certainly at least need to document this stuff better.  Seems like a subtle point that is easy to miss.  So there are const and non-const versions with the same function name and they do different things.  What is the use case for the const version, which may or may not give a valid answer and if in debug mode it may or may not crash the program?

isucan, line 781: The const and non-const version return the same object, but the non-const version also calls .update() on the RobotState. Both functions always give a valid answer at the time you call them, but for efficiency reasons FK is not computed all the time for every state update. This is why when you get the state, transforms could be dirty. The confusing thing I believe is that the assertion on dirty transforms is removed in release code -- again for efficiency reasons.I am thinking to actually have that check done every time though.

hersh, line 781: How about removing the non-const functions and making the transform data "mutable"?

Is there anything about the "cleaning" (i.e. calling update()) of the dirty transforms that logically changes a RobotState?

sachinchitta, line 781: One option is to rename them to be different and better reflect what's going on inside the function.

hersh, line 781: See my pull request: https://github.com/ros-planning/moveit_core/pull/106 and let me know what you think.

PR #370, opened 2013-11-18
Fix wrong cloud offset

YoheiKakiuchi:
For filtering points with sensor_range, distance between sensor_origin and a point was used for limiting range.
But, the point was not converted to the fixed frame, it still being on the sensor frame.
So, distance should be calculated at the sensor frame.


isucan:
This does not look correct to me. Can you explain what the problem is?

isucan:
i see this is superseded by #371

mikeferguson:
@isucan, this is not superceded by #371 -- 371 fixes hydro-devel, this targets groovy-devel.

YoheiKakiuchi:
groovy-devel can not be updated ?
Now, I am using groovy, I am happy to apply bug fix to debian package.

PR #371, opened 2013-11-18
Fix wrong cloud offset

YoheiKakiuchi:
For filtering points with sensor_range, distance between sensor_origin and a point was used for limiting range.
But, the point was not converted to the fixed frame, it still being on the sensor frame.
So, distance should be calculated at the sensor frame.


isucan:
Good catch! Thank you!

PR #373, opened 2013-11-22
Doxygen: added warnings and details to planning_scene_monitor.h

hersh:
Added warnings to PlanningSceneMonitor::getPlanningScene() functions and added a bunch of details to the docs for LockedPlanningSceneRO and RW.

@isucan especially, can you check over these comments for correctness?

### DIFF
--- a/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
+++ b/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
@@ -153,15 +153,28 @@ public:
     return robot_model_;
   }
 
-  /** @brief Get the planning scene
-   *  @return An instance of the planning scene*/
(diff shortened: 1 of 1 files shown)


planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
```diff
@@ -153,15 +153,28 @@ class PlanningSceneMonitor : private boost::noncopyable
     return robot_model_;
   }
 
-  /** @brief Get the planning scene
-   *  @return An instance of the planning scene*/
+  /** @brief <b>Avoid this function!</b>  Returns an @b
```
isucan, line 156: This pointer is unsafe only when there is a diff publishing thread running in the background (which is usually the case, when running in conjunction with rviz)

isucan:
This looks great to me! Thanks for adding these details!

PR #379, opened 2013-11-27
Re-ordered movegroup's initialization, so capabilities start after monitors.

### LINKED ISSUE #378: move_group: there is no way to tell when move_group is done initializing.
The move_group node takes a long time to start up, somewhere around 30 seconds for me.  I have written a "capability" for move_group which uses an action server.  I have written a client for the action server.

I have written some test launch files for my system, but they require a giant sleep() call, on the order of 40 seconds, to wait for move_group to finish initializing itself.  The amazing slowness of that initialization deserves its own bug, but that is not this bug.  This bug is about the fact that just sleeping 40 seconds is not a reliable way to wait for move_group to become ready.

A reliable way to wait for move_group to initialize itself would be to have move_group tell everyone else when it becomes ready.  Some options:
- move_group publishes a boolean topic called "/move_group/ready" or similar.
- move_group does its own initialization before initializing the capability plugins.
  - This would help me because my client can wait for my capability's action server to become ready with SimpleActionClient::waitForServer().

The second option is very easy, but I don't know if changing the order of initialization in move_group.cpp will break things.

hersh:
This is intended to fix #378.

Capabilities starting after monitors means that a client which waits on a capability to become ready (via SimpleActionClient::waitForServer() for instance) can be sure that the rest of move_group is already initialized and thus ready to be used.

In a long-running system this is not really an issue, but for tests which start up a new instance of move_group each time, this is a big problem.

I don't know much about the other capabilities for move_group, so I don't know if this will cause trouble for them.  I do know it makes my own particular capability and its client work nicely though.

### DIFF
--- a/move_group/src/move_group.cpp
+++ b/move_group/src/move_group.cpp
@@ -170,13 +170,13 @@ int main(int argc, char **argv)
     else
       ROS_INFO("MoveGroup debug mode is OFF");
 
-    move_group::MoveGroupExe mge(planning_scene_monitor, debug);
-
     printf(MOVEIT_CONSOLE_COLOR_CYAN "Starting context monitors...\n" MOVEIT_CONSOLE_COLOR_RESET);
     planning_scene_monitor->startSceneMonitor();
     planning_scene_monitor->startWorldGeometryMonitor();
     planning_scene_monitor->startStateMonitor();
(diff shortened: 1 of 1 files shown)


isucan:
This should be fine. +1

PR #381, opened 2013-11-29
add clear_octomap service

YoheiKakiuchi:
I added a service to force clearing octomap.
It needs to clear un-updated point when sensor see wrong points such as noise.


isucan:
This is somewhat inconsistent API. It affects only one plugin. We should have such functionality as part of the PlanningSceneMonitor. You can publish a PlanningScene diff (.diff set to true) that has an empty octomap. That will also achieve the functionality you desire.

PR #383, opened 2013-12-03
details about planPick() and planPlace() in doxygen comments

hersh:


manipulation/pick_place/include/moveit/pick_place/pick_place.h
```diff
@@ -141,11 +141,22 @@ class PickPlace : private boost::noncopyable,
...
-  /** \brief Plan the sequence of motions that perform a pickup action */
-  PickPlanPtr planPick(const planning_scene::PlanningSceneConstPtr &planning_scene, const moveit_msgs::PickupGoal &goal) const;
-
-  /** \brief Plan the sequence of motions that perform a placement action */
-  PlacePlanPtr planPlace(const planning_scene::PlanningSceneConstPtr &planning_scene, const moveit_msgs::PlaceGoal &goal) const;
+  /** @brief Plan the sequence of motions that perform a pickup action.
+   * @return a shared pointer to the resulting plan object.
+   *
+   * This function always either throws an exception or returns a plan
+   * pointer.  You don't need to check if the returned pointer is
```
isucan, line 148: Maybe we will change this in the future. I'd rather users always check pointers that could be null.

hersh:
The thing is, code that calls these functions is already looking like this:

```
  try
  {
    plan = pick_place_->planPick(scene, goal);
  }
  catch(std::runtime_error &ex)
  {
    ROS_ERROR("Pick&place threw an exception: %s", ex.what());
    return;
  }
  catch(...)
  {
    ROS_ERROR("Pick&place threw an exception");
    return;
  }
```

And after those 10 lines of failure handling code, I still need another 3 lines:

```
  if(plan)
  {
     // finally use the plan...
  }
```

Can we either commit to throwing exceptions to indicate failure or returning null to indicate failure, instead of doing both?

Actually, for this specific case, there is even yet still more failure-handling code required:

```
    std::vector<pick_place::ManipulationPlanPtr> plans = plan->getSuccessfulManipulationPlans();
    if(plans.size() == 0)
    {
        ROS_ERROR("Planning failed");
        return;
    }
```

So in the end, we could eliminate all the exceptions _and_ the possibility of a null pointer, and just return a PickPlanPtr that is always valid but sometimes contains successful plans and sometimes doesn't.

As it is, I've got a whole page of code and I've done a single function call.

sachinchitta:
We'll leave this as it is for now.

PR #385, opened 2013-12-08
comply to the new Table.msg

vrabaud:


isucan:
This looks good to me, but Sachin worked with this last.

PR #399, opened 2014-01-17
Fix: Parse error at "BOOST_JOIN"

bchretien:
See: https://bugreports.qt-project.org/browse/QTBUG-22829

This was tested on Arch Linux (Qt 4.8.5, Boost 1.55) while packaging for hydro.

### DIFF
--- a/benchmarks_gui/include/benchmark_processing_thread.h
+++ b/benchmarks_gui/include/benchmark_processing_thread.h
@@ -4,7 +4,9 @@
 #include <QtCore/QThread>
 #include <QProgressDialog>
 
+#ifndef Q_MOC_RUN
 #include <moveit/benchmarks/benchmark_execution.h>
+#endif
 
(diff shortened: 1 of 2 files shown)


bchretien:
I just found a similar error in `benchmarks_gui`, so wait for my amend before merging.

hersh:
Wow, you amended your single commit instead of just adding a second commit.  Interesting.  I wonder if Travis will notice the amended commit and re-run the build... I just checked and it looks like Travis took long enough to get started that it probably has the latest version.

Thanks for the patch!  I'll just wait until it passes our Travis build and then merge it.

bchretien:
@hersh: I didn't want to pollute the history with 2 commits, and normally Travis can handle that just fine. Anyhow, hydro's moveit_ros is now packaged for Arch Linux :-)

hersh:
Excellent, nice work.  Thanks for supporting moveit on Arch!

PR #401, opened 2014-01-19
add default_robot_unpadded_padded/default_robot_unpadded_scale to set pa...

k-okada:
...dding/scale for CollisionRobotUnpadded, see Issue #341

this patch requries https://github.com/ros-planning/moveit_core/pull/139

### DIFF
--- a/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
+++ b/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
@@ -401,6 +401,10 @@ protected:
   double                                default_robot_padd_;
   /// default robot scaling
   double                                default_robot_scale_;
+  /// default unpadded-robot padding
+  double                                default_robot_unpadded_padd_;
+  /// default unpadded-robot scaling
+  double                                default_robot_unpadded_scale_;
   /// default object padding
   double                                default_object_padd_;
   /// default attached padding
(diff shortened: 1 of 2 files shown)


isucan:
Hello Kei,

I am not sure I understand what problem this pull request fixes. By design, the unpadded version of the robot is meant to be the exact robot defined in the URDF, with no padding. Adding padding to that seems incorrect and I would rather not merge these changes. We can decide that we do not need unpadded robots at all, just different types of padding, but that can be done already, without removing the unpadded robot.

k-okada:
Hi, Ioan

The problem is that the self collision uses exact robot model defined in URDF, as seen in attached images, however we want get more 'safer' trajectory.  I hope that you agree on this point.

There might be several approaches. Are you suggesting to select 2) ?

1) add padding to unpadded robot

2) use padded robot for self collision

3) enable to choose type of collision robot 

screenshot_from_2014-01-20 10 31 52
screenshot_from_2014-01-20 10 31 46

isucan:
Hello Kei,

I absolutely agree we want safe paths. The reasoning in moveit right now is
that we use the padded version of the robot for collision checking at
panning time and we use the unpadded robot while we check the path online,
while it is executed. The idea is to be a bit less paranoid and not cancel
trajectories all the time during execution. Do you want to use the padded
robot for the checks during trajectory execution?
I think option 3) is a good one. If we have a parameter for changing the
behavior, that would be great.

We could also replace the unpadded robot idea with a separate padding for
online checking of trajectories. That may be a better way to go in the
future, but we need to rename variables and to check the other uses of the
unpadded robot - pick and place  mostly. It is starting to sound to me like
this would be the better option. What do other people think?

isucan:
This pull request can no longer be automatically merged and overlaps with #404 (which was merged). Please submit another pull request if more changes are needed.

PR #408, opened 2014-02-03
planning_scene_monitor: fix state update bug

acornacorn:
The rate of state updates is limited to dt_state_update_ per second.
When an update arrived it was not processed if another was recently
processed.  This meant that if a quick sequence of state updates
arrived and then no updates arrive for a while that the last update(s)
were not seen until another arrives (which may be much later or
never). This fixes the bug by periodically checking for pending
updates and running them if they have been pending longer than
dt_state_update_.

Fixes issue https://github.com/ros-planning/moveit_ros/issues/405 (replaces pull request https://github.com/ros-planning/moveit_ros/pull/407)

### DIFF
--- a/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
+++ b/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
@@ -310,7 +310,7 @@ public:
   /** @brief Get the topic names that the monitor is listening to */
   void getMonitoredTopics(std::vector<std::string> &topics) const;
 
-  /** \brief Return the time when the last update was made to the planning scene (by the monitor) */
+  /** \brief Return the time when the last update was made to the planning scene (by \e any monitor) */
   const ros::Time& getLastUpdateTime() const
   {
     return last_update_time_;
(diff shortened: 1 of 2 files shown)


planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
```diff
@@ -450,15 +450,34 @@ class PlanningSceneMonitor : private boost::noncopyable
...
 
-  /// the planning scene state is updated at a maximum specified frequency,
-  /// and this timestamp is used to implement that functionality
+
+  // Lock for state_update_pending_ and dt_state_update_
+  boost::mutex state_pending_mutex_;
+
+  /// True when we need to update the RobotState from current_state_monitor_
+  // This field is protected by state_pending_mutex_
+  volatile bool state_update_pending_;
```
isucan, line 468: The volatile keyword sometimes does different things on different architectures. I am not sure we even need it.

isucan:
Excellent fix! +1

shaun-edwards:
Verified the fix on my system...+1

PR #409, opened 2014-02-03
Fix planning scene display initial state

acornacorn:
When rviz starts after move_group, the planning scene diff messages sent by move_group do not give PlanningSceneDisplay the full state of the planning scene.  For example if the robot has not moved for a while then move_group may not be sending updates of the robot state and the state of the PlanningScene robot will be different from the actual state of the real robot (it will appear in the default position rather than the current position).

This change fixes the PlanningSceneDisplay so that when it starts up (robot model is loaded or planning scene monitor topic changes) it will query move_group for the current full PlanningScene state and use that to initialize the PlanningScene maintained by the PlanningSceneDisplay.

Fixes issue https://github.com/ros-planning/moveit_ros/issues/405
Replaces https://github.com/ros-planning/moveit_ros/pull/407 and adds to https://github.com/ros-planning/moveit_ros/pull/408

### DIFF
--- a/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
+++ b/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
@@ -310,7 +310,7 @@ public:
   /** @brief Get the topic names that the monitor is listening to */
   void getMonitoredTopics(std::vector<std::string> &topics) const;
 
-  /** \brief Return the time when the last update was made to the planning scene (by the monitor) */
+  /** \brief Return the time when the last update was made to the planning scene (by \e any monitor) */
   const ros::Time& getLastUpdateTime() const
   {
     return last_update_time_;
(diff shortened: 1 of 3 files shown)


planning/planning_scene_monitor/src/planning_scene_monitor.cpp
```diff
@@ -386,14 +394,52 @@ bool planning_scene_monitor::PlanningSceneMonitor::updatesScene(const planning_s
...
+
+  if (client.call(srv))
+  {
+    newPlanningSceneMessage(srv.response.scene);
+  }
+  else
+  {
+    ROS_ERROR("Failed to call service %s at %s:%d",
+      service_name.c_str(),
+      __FILE__,
```
isucan, line 430: **FILE** and **LINE** are already included in the call to ROS_ERROR. You can see these in either rosconsole, or you can configure your rosconsole output with ROSCONSOLE_FORMAT (http://wiki.ros.org/rosconsole)

planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
```diff
@@ -92,6 +92,9 @@ class PlanningSceneMonitor : private boost::noncopyable
   /// The name of the topic used by default for receiving full planning scenes or planning scene diffs
   static const std::string DEFAULT_PLANNING_SCENE_TOPIC; // "/planning_scene"
   
+  /// The name of the service used by default for requesting full planning scene state
+  static const std::string DEFAULT_PLANNING_SCENE_SERVICE; // "/get_planning_scene"
```
isucan, line 96: This is very good, but there is a duplicate name now in capability_names.h, in move_group. It would be great if we could keep the name here, but have capability_names.h define `GET_PLANNING_SCENE_SERVICE_NAME` to be set to `PlanningSceneMonitor::DEFAULT_PLANNING_SCENE_SERVICE`. We would need to move the strings names from capability_names.h into a .cpp file that gets linked into the `moveit_move_group_capabilities_base` library.

isucan, line 96: The other option is to have no default value for the argument, and then have the rviz plugin pass the value it reads from capability_names.h; that may be a cleaner solution in terms of deps? Then again, it is nice to have the function work with nor args too..

isucan:
+1

planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
acornacorn, line 96: Perhaps we should have these names in a more accessible location.  My initial thought is a header file in the moveit_msgs package, but I think that would require the names to be #defines (which is fine with me).

If you don't like #defines then another option is a (new) library dedicated to names somewhere in moveit_ros.

sachinchitta:
+1

Just tested this with a hacked version of the industrial robot simulator that publishes initial values to all be 1.0. It works but requires an additional line in move_group.launch to configure the capability for getting planning scene - here's the full param for capabilities in move_group.launch. The additional line is the last one that adds move_group/MoveGroupGetPlanningSceneService

```
<param name="capabilities" value="move_group/MoveGroupCartesianPathService
                  move_group/MoveGroupExecuteService
                  move_group/MoveGroupKinematicsService
                  move_group/MoveGroupMoveAction
                  move_group/MoveGroupPickPlaceAction
                  move_group/MoveGroupPlanService
                  move_group/MoveGroupQueryPlannersService
                  move_group/MoveGroupStateValidationService
                  move_group/MoveGroupGetPlanningSceneService
                  " />
```

mikeferguson:
+1 as well. We should update the setup assistant to add the GetPlanningSceneService plugin though  (I've manually added it in the past for our robot). Should probably also send an email to the mailing list when this gets merged/released so that people know to update their already exported configurations.

PR #410, opened 2014-02-03
ApproachAndTranslateStage dynamic reconfigure bug fixed.

hersh:
The bug shows up in test code, where it becomes apparent that creating a ros::NodeHandle
in a static initializer makes it very difficult to call ros::init() before creating
the first NodeHandle.


isucan:
Looks good! This must have been tough to find! Thanks!
In general for fixing this problem MoveIt typically uses a static var local to a function,
so something like:

```
MyInstance& GetInstance()
{
   static MyInstance instance;
   return instance;
}
```

This approach avoid the memory leak that you get with constructing the object yourself.

acornacorn:
+1

PR #413, opened 2014-02-07
adding get for active joints

sachinchitta:


isucan:
+1

adolfo-rt:
@sachinchitta. Thanks for this patch, really useful. I'm already using it as of today :-). I see that the last release was just two days before this got merged, but I'm still going to ask. Is there a patch release scheduled for the near future?.

PR #416, opened 2014-02-12
Separate InteractionHandler from RobotInteraction

acornacorn:
This splits the InteractionHandler class into its own files separate from RobotInteraction.

There are also some additional comments and moving some members from protected to private.

This is a first step in an attempt to simplify RobotInteraction and fix some threading issues that I have been causing me trouble.

The Ogre related changes are in preparation for new versions of Ogre (makes the build work with Ogre 1.7, 1.8, 1.9 and 1.10)


robot_interaction/include/moveit/robot_interaction/robot_interaction.h
```diff
@@ -34,8 +34,8 @@
 
 /* Author: Ioan Sucan, Adam Leeper */
 
-#ifndef MOVEIT_ROBOT_INTERACTION_ROBOT_INTERACTION_
-#define MOVEIT_ROBOT_INTERACTION_ROBOT_INTERACTION_
+#ifndef MOVEIT__ROBOT_INTERACTION__ROBOT_INTERACTION_H
```
isucan: This is not consistent with the rest of moveit, but if you want to change the convention, that is fine by me (as long as we are consistent)

robot_interaction/include/moveit/robot_interaction/interaction_handler.h
```diff
@@ -0,0 +1,330 @@
...
+  {
+    return kinematics_query_options_;
+  }
+
+  void setKinematicsQueryOptions(const kinematics::KinematicsQueryOptions &opt)
+  {
+    kinematics_query_options_ = opt;
+  }
+
+  void setKinematicsQueryOptionsForGroup(const std::string& group_name,
```
isucan, line 125: Unrelated note, but so I don't forget to mention this:

I don't remember if it was me that added these functions or not, but I think they are a bad thing to have. We should refactor and try to avoid them.

isucan:
This looks great! ++1

acornacorn:
Travis build choked on my Ogre related changes.  Looking into it...

robot_interaction/include/moveit/robot_interaction/interaction_handler.h
acornacorn, line 125: You're right: I should stick to the convention.  Fixed.

PR #422, opened 2014-02-15
Improve RobotInteraction thread safety

acornacorn:
This change attempts to simplify RobotInteraction and improve thread safety.

These changes focus on InteractionHandler.  Some of the functionality is pulled out into helper classes (LockedRobotState and KinematicOptionsMap) so that the intent of InteractionHandler is more clear.

All non-const members are now protected with internal locks.
The locking scheme is documented in comments.

The state update is simplified, removing the need for condition variables.  Instead in-place updates are done using callback functions with the modifyState method (inherited by RobotInteraction from LockedRobotState).

### DIFF
--- a/robot_interaction/include/moveit/robot_interaction/interaction.h
+++ b/robot_interaction/include/moveit/robot_interaction/interaction.h
@@ -113,8 +113,8 @@ typedef boost::function<bool(
 ///           is somehow invalid or erronious (e.g. in collision).  true if
 ///           everything worked well.
 typedef boost::function<bool(
-	  robot_state::RobotState& state,
-	  const visualization_msgs::InteractiveMarkerFeedbackConstPtr & feedback)>
+    robot_state::RobotState& state,
+    const visualization_msgs::InteractiveMarkerFeedbackConstPtr & feedback)>
                                                 ProcessFeedbackFn;
 
 /// Type of function for updating marker pose for new state.
--- a/robot_interaction/include/moveit/robot_interaction/interaction.h
+++ b/robot_interaction/include/moveit/robot_interaction/interaction.h
@@ -1,0 +1,186 @@
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2014, SRI International
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of Willow Garage nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Acorn Pooley */
+
+#ifndef MOVEIT_ROBOT_INTERACTION_INTERACTION_
+#define MOVEIT_ROBOT_INTERACTION_INTERACTION_
+
+#include <visualization_msgs/InteractiveMarkerFeedback.h>
+#include <visualization_msgs/InteractiveMarker.h>
+#include <interactive_markers/menu_handler.h>
+#include <moveit/robot_state/robot_state.h>
+#include <boost/function.hpp>
+#include <boost/thread.hpp>
+#include <tf/tf.h>
+
+namespace moveit
+{
+namespace core
+{
+class RobotState;
+}
+}
+
+namespace robot_interaction
+{
+
+namespace InteractionStyle
+{
+  /// The different types of interaction that can be constructed for an end
+  /// effector This is a bitmask so OR together the parts you want.
+  enum InteractionStyle
+  {
+    POSITION_ARROWS = 1,      // arrows to change position
+    ORIENTATION_CIRCLES = 2,  // circles to change orientation
+    POSITION_SPHERE = 4,      // sphere: drag to change position
+    ORIENTATION_SPHERE = 8,   // sphere: drag to change orientation
+    POSITION_EEF = 16,        // drag end effector to change position
+    ORIENTATION_EEF = 32,     // drag end effector to change orientation
+    FIXED = 64,               // keep arrow and circle axis fixed
+
+    POSITION = POSITION_ARROWS |
+               POSITION_SPHERE |
+               POSITION_EEF,
+    ORIENTATION = ORIENTATION_CIRCLES |
+                  ORIENTATION_SPHERE |
+                  ORIENTATION_EEF,
+    SIX_DOF = POSITION |
+           ORIENTATION,
+    SIX_DOF_SPHERE = POSITION_SPHERE |
+                  ORIENTATION_SPHERE,
+    POSITION_NOSPHERE = POSITION_ARROWS |
+                        POSITION_EEF,
+    ORIENTATION_NOSPHERE = ORIENTATION_CIRCLES |
+                           ORIENTATION_EEF,
+    SIX_DOF_NOSPHERE = POSITION_NOSPHERE |
+                    ORIENTATION_NOSPHERE
+  };
+}
+
+/// Type of function for constructing markers.
+/// This callback sets up the marker used for an interaction.
+typedef boost::function<bool(
+          const robot_state::RobotState&,
+          visualization_msgs::InteractiveMarker&)>
+                                                InteractiveMarkerConstructorFn;
+
+/// Type of function for processing marker feedback.
+/// This callback function handles feedback for an Interaction's marker.
+/// Callback should update the robot state that was passed in according to
+/// the new position of the marker. Return true if the update was successful.
+/// Return false if the state was not successfully updated.
+typedef boost::function<bool(
+          robot_state::RobotState&,
+          const visualization_msgs::InteractiveMarkerFeedbackConstPtr &)>
+                                                ProcessFeedbackFn;
+
+/// Type of function for updating marker pose for new state.
+/// This callback is called when the robot state has changed.
+/// Callback should calculate a new
+/// pose for the marker based on the passed in robot state.
+/// Return true if the pose was modified, false if no update is needed (i.e. if
+/// the pose did not change).
+typedef boost::function<bool(
+          const robot_state::RobotState&,
+          geometry_msgs::Pose&)>
+                                                InteractiveMarkerUpdateFn;
+
+/// Representation of a generic interaction.
+/// Displays one interactive marker.
+struct GenericInteraction
+{
+  // Callback to construct interactive marker.
+  // See comment on typedef above.
+  InteractiveMarkerConstructorFn construct_marker;
+
+  // Callback to handle interactive marker feedback messages.
+  // See comment on typedef above.
+  ProcessFeedbackFn process_feedback;
+
+  // Callback to update marker pose when RobotState changes.
+  // See comment on typedef above.
+  InteractiveMarkerUpdateFn update_pose;
+
+  // Suffix added to name of markers.
+  // Automatically generated.
+  std::string marker_name_suffix;
+};
+
+/// Representation of an interaction via an end-effector
+/// Displays one marker for manipulating the EEF position.
+class EndEffectorInteraction
+{
+public:
+  /// The name of the group that sustains the end-effector (usually an arm)
+  std::string parent_group;
+
+  /// The name of the link in the parent group that connects to the end-effector
+  std::string parent_link;
+
+  /// The name of the group that defines the group joints
+  std::string eef_group;
+
+  /// Which degrees of freedom to enable for the end-effector
+  InteractionStyle::InteractionStyle interaction;
+
+  /// The size of the end effector group (diameter of enclosing sphere)
+  double size;
+
+};
+
+/// Representation of a joint interaction.
+/// Displays one marker for manipulating the joint.
+class JointInteraction
+{
+public:
+  /// The link in the robot model this joint is a parent of
+  std::string connecting_link;
+
+  /// The name of the frame that is a parent of this joint
+  std::string parent_frame;
+
+  /// The name of the joint
+  std::string joint_name;
+
+  /// The type of joint disguised as the number of DOF it has.  3=PLANAR in X/Y; 6=FLOATING
+  unsigned int dof;
+
+  /// The size of the connecting link  (diameter of enclosing sphere)
+  double size;
+};
+
+}
+
+#endif
--- a/robot_interaction/include/moveit/robot_interaction/interaction.h
+++ b/robot_interaction/include/moveit/robot_interaction/interaction.h
@@ -91,27 +91,41 @@ namespace InteractionStyle
 
 /// Type of function for constructing markers.
 /// This callback sets up the marker used for an interaction.
+/// @param state the current state of the robot
+/// @param marker the function should fill this in with an InteractiveMarker
+///          that will be used to control the interaction.
+///  @returns true if the function succeeds, false if the function was not able
+///          to fill in \e marker.
 typedef boost::function<bool(
-          const robot_state::RobotState&,
-          visualization_msgs::InteractiveMarker&)>
+          const robot_state::RobotState& state,
+          visualization_msgs::InteractiveMarker& marker)>
                                                 InteractiveMarkerConstructorFn;
 
 /// Type of function for processing marker feedback.
 /// This callback function handles feedback for an Interaction's marker.
 /// Callback should update the robot state that was passed in according to
-/// the new position of the marker. Return true if the update was successful.
-/// Return false if the state was not successfully updated.
+/// the new position of the marker.
+///
+/// @param state the current state of the robot
+/// @param marker the function should fill this in with an InteractiveMarker
+///          that will be used to control the interaction.
+/// @returns false if the state was not successfully updated or the new state
+///           is somehow invalid or erronious (e.g. in collision).  true if
+///           everything worked well.
 typedef boost::function<bool(
-          robot_state::RobotState&,
-          const visualization_msgs::InteractiveMarkerFeedbackConstPtr &)>
+	  robot_state::RobotState& state,
+	  const visualization_msgs::InteractiveMarkerFeedbackConstPtr & feedback)>
                                                 ProcessFeedbackFn;
 
 /// Type of function for updating marker pose for new state.
 /// This callback is called when the robot state has changed.
 /// Callback should calculate a new
 /// pose for the marker based on the passed in robot state.
-/// Return true if the pose was modified, false if no update is needed (i.e. if
-/// the pose did not change).
+/// @param state the new state of the robot
+/// @param pose the function should fill this in with the new pose of the
+///              marker, given the new state of the robot.
+/// @returns true if the pose was modified, false if no update is needed (i.e.
+///              if the pose did not change).
 typedef boost::function<bool(
           const robot_state::RobotState&,
           geometry_msgs::Pose&)>
@@ -140,9 +154,8 @@ struct GenericInteraction
 
 /// Representation of an interaction via an end-effector
 /// Displays one marker for manipulating the EEF position.
-class EndEffectorInteraction
+struct EndEffectorInteraction
 {
-public:
   /// The name of the group that sustains the end-effector (usually an arm)
   std::string parent_group;
 
@@ -162,9 +175,8 @@ public:
 
 /// Representation of a joint interaction.
 /// Displays one marker for manipulating the joint.
-class JointInteraction
+struct JointInteraction
 {
-public:
   /// The link in the robot model this joint is a parent of
   std::string connecting_link;
 
--- a/robot_interaction/include/moveit/robot_interaction/interaction_handler.h
+++ b/robot_interaction/include/moveit/robot_interaction/interaction_handler.h
@@ -346,7 +346,7 @@ public:
   const kinematics::KinematicsQueryOptions& getKinematicsQueryOptions() const;
   void setKinematicsQueryOptions(const kinematics::KinematicsQueryOptions &opt);
   void setKinematicsQueryOptionsForGroup(const std::string& group_name, 
-           kinematics::KinematicsQueryOptions &options);
+           const kinematics::KinematicsQueryOptions &options);
 };
 
 
--- a/robot_interaction/include/moveit/robot_interaction/interaction_handler.h
+++ b/robot_interaction/include/moveit/robot_interaction/interaction_handler.h
@@ -62,11 +62,11 @@ class GenericInteraction;
 
 
 /// Function type for notifying client of RobotState changes.
-//
-/// This callback function is called by the InteractionHandler::handle* functions,
-/// when changes are made to the internal robot state the handler maintains.
-/// The handler passes its own pointer as argument to the callback, as well
-/// as a boolean flag that indicates whether the error state changed --
+///
+/// This callback function is called by the InteractionHandler::handle*
+/// functions, when changes are made to the internal robot state the handler
+/// maintains.  The handler passes its own pointer as argument to the callback,
+/// as well as a boolean flag that indicates whether the error state changed --
 /// whether updates to the robot state performed in the
 /// InteractionHandler::handle* functions have switched from failing to
 /// succeeding or the other way around.
@@ -80,7 +80,8 @@ typedef boost::function<void(InteractionHandler*, bool)> InteractionHandlerCallb
 /// various joints in one group of one RobotState.
 /// The group being controlled is maintained by the RobotInteraction object
 /// that contains this InteractionHandler object.
-/// All InteractionHandler objects in the same RobotInteraction are controlling the same group.
+/// All InteractionHandler objects in the same RobotInteraction are controlling
+/// the same group.
 class InteractionHandler : public LockedRobotState
 {
 public:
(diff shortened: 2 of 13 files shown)


robot_interaction/include/moveit/robot_interaction/interaction.h
```diff
@@ -0,0 +1,186 @@
...
+    ORIENTATION_NOSPHERE = ORIENTATION_CIRCLES |
+                           ORIENTATION_EEF,
+    SIX_DOF_NOSPHERE = POSITION_NOSPHERE |
+                    ORIENTATION_NOSPHERE
+  };
+}
+
+/// Type of function for constructing markers.
+/// This callback sets up the marker used for an interaction.
+typedef boost::function<bool(
```
isucan, line 99: What does the callback return? How are the arguments used?

robot_interaction/include/moveit/robot_interaction/interaction_handler.h
```diff
@@ -37,11 +37,43 @@
...
+//#include <moveit/robot_interaction/robot_interaction.h>
+#include <visualization_msgs/InteractiveMarkerFeedback.h>
+#include <interactive_markers/menu_handler.h>
+#include <tf/tf.h>
 
 namespace robot_interaction
 {
 
+class InteractionHandler;
+typedef boost::shared_ptr<InteractionHandler> InteractionHandlerPtr;
```
isucan, line 50: why not use MOVEIT_CLASS_FORWARD from macros/?

robot_interaction/include/moveit/robot_interaction/interaction.h
```diff
@@ -0,0 +1,186 @@
...
+                                                InteractiveMarkerConstructorFn;
+
+/// Type of function for processing marker feedback.
+/// This callback function handles feedback for an Interaction's marker.
+/// Callback should update the robot state that was passed in according to
+/// the new position of the marker. Return true if the update was successful.
+/// Return false if the state was not successfully updated.
+typedef boost::function<bool(
+          robot_state::RobotState&,
+          const visualization_msgs::InteractiveMarkerFeedbackConstPtr &)>
```
isucan: Why take a ConstPtr& here for a ROS message and just a & for the callback above? I'd rather we take the ref to the msg in both cases.

robot_interaction/include/moveit/robot_interaction/interaction.h
```diff
@@ -0,0 +1,186 @@
...
+  std::string parent_link;
+
+  /// The name of the group that defines the group joints
+  std::string eef_group;
+
+  /// Which degrees of freedom to enable for the end-effector
+  InteractionStyle::InteractionStyle interaction;
+
+  /// The size of the end effector group (diameter of enclosing sphere)
+  double size;
```
isucan, line 172: Should we have a default constructor that sets `size` and `interaction`?

robot_interaction/include/moveit/robot_interaction/interaction.h
```diff
@@ -0,0 +1,186 @@
...
+  std::string parent_frame;
+
+  /// The name of the joint
+  std::string joint_name;
+
+  /// The type of joint disguised as the number of DOF it has.  3=PLANAR in X/Y; 6=FLOATING
+  unsigned int dof;
+
+  /// The size of the connecting link  (diameter of enclosing sphere)
+  double size;
```
isucan, line 193: Constructor for dof and size?

robot_interaction/include/moveit/robot_interaction/interaction.h
```diff
@@ -0,0 +1,186 @@
...
+  InteractiveMarkerUpdateFn update_pose;
+
+  // Suffix added to name of markers.
+  // Automatically generated.
+  std::string marker_name_suffix;
+};
+
+/// Representation of an interaction via an end-effector
+/// Displays one marker for manipulating the EEF position.
+class EndEffectorInteraction
```
isucan: In general the convention in MoveIt is to use structs when all members are public. The object above is a struct. Could we make this one and below structs as well?

robot_interaction/include/moveit/robot_interaction/locked_robot_state.h
```diff
@@ -0,0 +1,116 @@
...
+// be modified, a copy is made and the copy is modified.  Any externally held
+// references will be out of date but still valid.
+//
+// The RobotState can only be modified by passing a callback function which
+// does the modification.
+class LockedRobotState
+{
+public:
+  LockedRobotState(const robot_state::RobotStatePtr& state);
+  LockedRobotState(const robot_state::RobotState& state);
```
isucan, line 62: Do we need all these constructors? If not, I propose removing one of the two above. My concern is that the semantics is slightly different. If I use one, then I modify the state I passed to LockedRobotState, for the first case, the change is visible in this object, and for the second, the change is not visible. I would prefer to keep just the plain & constructor, but not the RobotStatePtr& constructor.

robot_interaction/include/moveit/robot_interaction/locked_robot_state.h
```diff
@@ -0,0 +1,116 @@
...
+  // This is a function that can modify the maintained state.
+  typedef boost::function<void (robot_state::RobotState*)> ModifyStateFunction;
+  
+  // Modify the state.
+  //
+  // This modifies the state by calling \e modify on it.
+  // The \e modify function is passed a reference to the state which it can
+  // modify.  No threads will be given access to the state while the \e modify
+  // function is running.
+  void modifyState(ModifyStateFunction modify);
```
isucan: can the argument be const &?

robot_interaction/include/moveit/robot_interaction/locked_robot_state.h
```diff
@@ -0,0 +1,116 @@
...
+
+protected:
+  // This is called when the internally maintained state has changed.
+  // This is called with state_lock_ unlocked.
+  // Default definition does nothing.  Override to get notification of state
+  // change.
+  // TODO: is this needed?
+  virtual void robotStateChanged();
```
isucan, line 96: Can't hurt. I'd say keep it.

robot_interaction/include/moveit/robot_interaction/locked_robot_state.h
```diff
@@ -0,0 +1,116 @@
...
+
+protected:
+  // this locks all accesses to the state_ member.
+  // The lock can also be used by subclasses to lock additional fields.
+  mutable boost::mutex state_lock_;
+
+private:
```
isucan, line 103: I think state_ should be protected as well.

robot_interaction/include/moveit/robot_interaction/robot_interaction.h
```diff
@@ -39,12 +39,12 @@
 
 #include <visualization_msgs/InteractiveMarkerFeedback.h>
 #include <visualization_msgs/InteractiveMarker.h>
-#include <interactive_markers/menu_handler.h>
+//#include <interactive_markers/menu_handler.h>
 #include <moveit/robot_state/robot_state.h>
-#include <moveit/macros/class_forward.h>
```
isucan, line 44: You seem to not like this... Any reason why? :)

robot_interaction/include/moveit/robot_interaction/robot_interaction.h
```diff
@@ -211,38 +120,51 @@ class RobotInteraction
...
     return active_vj_;
   }
 
-  static bool updateState(robot_state::RobotState &state, const EndEffector &eef, const geometry_msgs::Pose &pose,
-                          unsigned int attempts, double ik_timeout,
+  const robot_model::RobotModelConstPtr& getRobotModel() const { robot_model_; }
+
+  // Get the kinematic options map.
+  // Use this to set kinematic options (defaults or per-group).
+  KinematicOptionsMapPtr getKinematicOptionsMap() { return kinematic_options_map_; }
```
isucan, line 165: can we return const&?

robot_interaction/include/moveit/robot_interaction/robot_interaction.h
```diff
@@ -211,38 +120,51 @@ class RobotInteraction
...
+                          const EndEffectorInteraction &eef,
+                          const geometry_msgs::Pose &pose,
+                          unsigned int attempts,
+                          double ik_timeout,
                           const robot_state::GroupStateValidityCallbackFn &validity_callback = robot_state::GroupStateValidityCallbackFn(),
                           const kinematics::KinematicsQueryOptions &kinematics_query_options = kinematics::KinematicsQueryOptions());
-  static bool updateState(robot_state::RobotState &state, const Joint &vj, const geometry_msgs::Pose &pose);
+
+
+#if 0
```
isucan: Maybe just delete the disabled code?

robot_interaction/src/interaction_handler.cpp
```diff
@@ -54,69 +56,108 @@
...
+,tf_(tf)
+,display_meshes_(true)
+,display_controls_(true)
+,kinematic_options_map_(new KinematicOptionsMap)
+,robot_interaction_(NULL)
+,planning_frame_(robot_model->getModelFrame())
+{
+}
+
+std::string InteractionHandler::fixName(std::string name)
```
isucan, line 123: This should probably just be a function in an anonymous namespace.

robot_interaction/include/moveit/robot_interaction/interaction_handler.h
```diff
@@ -315,16 +295,61 @@ class RobotInteraction::InteractionHandler
...
+public:
+  // DEPRECATED FUNCTIONS.
+  // DO NOT USE THESE.  Instead access the KinematicOptions by calling RobotInteraction::getKinematicOptionsMap()
+  void setGroupStateValidityCallback(const robot_state::GroupStateValidityCallbackFn &callback);
+  void setIKTimeout(double timeout);
+  void setIKAttempts(unsigned int attempts);
+  const kinematics::KinematicsQueryOptions& getKinematicsQueryOptions() const;
+  void setKinematicsQueryOptions(const kinematics::KinematicsQueryOptions &opt);
+  void setKinematicsQueryOptionsForGroup(const std::string& group_name, 
+           kinematics::KinematicsQueryOptions &options);
```
isucan: const options?

robot_interaction/src/interaction_handler.cpp
```diff
@@ -332,47 +319,152 @@ void RobotInteraction::InteractionHandler::handleJoint(const robot_interaction::
...
+// MUST hold state_lock_ when calling this!
+void InteractionHandler::updateStateEndEffector(
+      robot_state::RobotState* state,
+      const EndEffectorInteraction* eef,
+      const geometry_msgs::Pose* pose,
+      StateChangeCallbackFn *callback)
+{
+  // This is called with state_lock_ held, so no additional locking needed to
+  // access kinematic_options_map_.
+  KinematicOptions kinematic_options = kinematic_options_map_->getOptions(eef->parent_group);
```
isucan, line 361: const& for return type?

robot_interaction/src/interaction_handler.cpp
```diff
@@ -407,4 +499,156 @@ bool RobotInteraction::InteractionHandler::transformFeedbackPose(const visualiza
...
+  //
+  // This is a nop if a constructor with a robot_interaction parameter is used.
+  shared_kinematic_options_map->merge(*kinematic_options_map_);
+
+  // from now on the InteractionHandler shares the same KinematicOptionsMap
+  // with RobotInteraction.
+  // The old *kinematic_options_map_ is automatically deleted by boost::shared_ptr.
+  //
+  // This is a nop if a constructor with a robot_interaction parameter is used.
+  kinematic_options_map_ = shared_kinematic_options_map;
```
isucan, line 546: You could actually just call .swap() here.

robot_interaction/include/moveit/robot_interaction/kinematic_options.h
```diff
@@ -0,0 +1,106 @@
...
+namespace robot_interaction
+{
+
+// Options for inverse kinematics calculations.
+//
+// This is intended to be lightweight and passable by value.  No virtual
+// functions and no destructor.
+struct KinematicOptions
+{
+public:
```
isucan: No need for `public:` in struct.

robot_interaction/include/moveit/robot_interaction/kinematic_options.h
```diff
@@ -0,0 +1,106 @@
...
+  };
+
+
+  /// Set \e state using inverse kinematics
+  /// @param state the state to set
+  /// @param group name of group whose joints can move
+  /// @param tip link that will be posed
+  /// @param pose desired pose of tip link
+  /// @param result true if IK succeeded.
+  void setStateFromIK(robot_state::RobotState* state,
```
isucan: I am all for pointers, but this is slightly inconsistent with the rest of the components in the system. Perhaps I have not seen all the call sites, but the function internally assumes none of the values are nullptr. I think we should set all but `state` to be const & and result should be a return value of the function.

robot_interaction/src/kinematic_options.cpp
```diff
@@ -0,0 +1,132 @@
...
+// This is intended to be called as a ModifyStateFunction to modify the state
+// maintained by a LockedRobotState in place.
+void robot_interaction::KinematicOptions::setStateFromIK(
+      robot_state::RobotState* state,
+      const std::string* group,
+      const std::string* tip,
+      const geometry_msgs::Pose* pose,
+      bool* result) const
+{
+  const robot_model::JointModelGroup *jmg = state->getJointModelGroup(*group);
```
isucan: We should check that jmg is not nullptr and return false if it is.

robot_interaction/include/moveit/robot_interaction/interaction_handler.h
```diff
@@ -213,89 +181,101 @@ class RobotInteraction::InteractionHandler
...
 
-  /** \brief Clear any error settings. This makes the markers appear as if the state is no longer invalid. */
+  /** \brief Clear any error settings.
+   * This makes the markers appear as if the state is no longer invalid. */
   void clearError(void);
 
+  /** \brief This should only be called by RobotInteraction.
+   * Associates this InteractionHandler to a RobotInteraction.
+   */
+  void setRobotInteraction(RobotInteraction* robot_interaction);
```
isucan, line 244: I don't think this should be a public function.

isucan, line 244: If you really need to, I think friend functions as a last resort are better than the public function, but I think that we can avoid even that.

robot_interaction/include/moveit/robot_interaction/kinematic_options_map.h
```diff
@@ -0,0 +1,116 @@
...
+#include <moveit/robot_interaction/kinematic_options.h>
+#include <boost/thread.hpp>
+#include <boost/function.hpp>
+
+namespace robot_interaction
+{
+
+// Maintains a set of KinematicOptions with a key/value mapping and a default
+// value.
+struct KinematicOptionsMap
```
isucan: This should be a class.

robot_interaction/include/moveit/robot_interaction/kinematic_options_map.h
```diff
@@ -0,0 +1,116 @@
...
+  static const std::string ALL;
+
+  /// Set \e state using inverse kinematics.
+  /// @param state the state to set
+  /// @param key used to lookup the options to use
+  /// @param group name of group whose joints can move
+  /// @param tip link that will be posed
+  /// @param pose desired pose of tip link
+  /// @param result true if IK succeeded.
+  void setStateFromIK(robot_state::RobotState* state,
```
isucan: Same comment as previously. key, group, tip, pose should be const&. There is no need to state to be ptr either (for consistency reasons), although I do like having a ptr so it is apparent to the caller the arg is modified. Result should be returned by value. Please please please make this change :) Pretty please.

robot_interaction/src/interaction_handler.cpp
```diff
@@ -47,6 +48,7 @@
 #include <boost/math/constants/constants.hpp>
 #include <algorithm>
 #include <limits>
+#include <stdexcept>
```
isucan: Why do we include this if we don't throw exceptions? Also, we should probably use moveit/exceptions/

robot_interaction/src/robot_interaction.cpp
```diff
@@ -614,8 +641,10 @@ bool RobotInteraction::updateState(robot_state::RobotState &state, const Joint &
   state.update();
   return true;
 }
+#endif
 
-bool RobotInteraction::updateState(robot_state::RobotState &state, const EndEffector &eef, const geometry_msgs::Pose &pose,
+// TODO: can we get rid of this?  Only used in moveit_ros/benchmarks_gui/src/tab_states_and_goals.cpp right now.
+bool RobotInteraction::updateState(robot_state::RobotState &state, const EndEffectorInteraction &eef, const geometry_msgs::Pose &pose,
```
isucan: +1 for getting rid of it.

robot_interaction/src/robot_interaction.cpp
```diff
@@ -178,7 +185,7 @@ void RobotInteraction::decideActiveJoints(const std::string &group)
       {
         if (vj[i].type_ == "planar" || vj[i].type_ == "floating")
         {
-          Joint v;
+          JointInteraction v;
```
isucan, line 192: +100 for these renames! Thank you!

robot_interaction/src/kinematic_options.cpp
```diff
@@ -0,0 +1,132 @@
...
+      OptionBitmask fields)
+{
+  // This function is implemented with the O_FIELDS and QO_FIELDS macros to
+  // ensure that any fields added to robot_interaction::KinematicOptions or
+  // kinematics::KinematicsQueryOptions are also added here and to the
+  // KinematicOptions::OptionBitmask enum.
+
+  // This needs to represent all the fields in
+  // robot_interaction::KinematicOptions except options_
+  #define O_FIELDS(F) \
```
isucan, line 83: Normally I'd love running this in my head to see what is going on but I wanna watch a movie
+1 for coolness :)

robot_interaction/src/kinematic_options_map.cpp
```diff
@@ -0,0 +1,154 @@
...
+      const KinematicOptionsMap& other)
+{
+  if (&other == this)
+    return;
+
+  // need to lock in consitent order to avoid deadlock. 
+  // Lock the one with lower address first.
+  boost::mutex *m1 = &lock_;
+  boost::mutex *m2 = &other.lock_;
+  if (m2 < m1)
```
isucan, line 126: cool!

robot_interaction/src/kinematic_options_map.cpp
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// merge other into this.  All options in other take precedence over this.
+void robot_interaction::KinematicOptionsMap::merge(
+      const KinematicOptionsMap& other)
+{
+  if (&other == this)
+    return;
+
+  // need to lock in consitent order to avoid deadlock.
```
isucan: typo

isucan:
+1 for this change. There are some comments that I'd really like to see addressed (setStateFromIK, for instance) and I had some questions in a few places, but this is a very cool improvement!  Thanks!

robot_interaction/include/moveit/robot_interaction/interaction_handler.h
acornacorn, line 244: I really wanted to make this a private function, but to do so would mean RobotInteraction would have to be a friend of InteractionHandler, which I really don't want to do (breaks encapsulation).  I considered adding a new class which is a friend of InteractionHandler and is inherited by RobotInteraction with a single method that calls setRobotInteraction, but that seemed like a lot of complexity.

On the other hand, making this a public function does not really hurt anything.  It is an error (print error and do nothing) to call it with anything other than the RobotInteraction it is already associated with, and calling it with that RobotInteraction does nothing.  So it is benign to call this.  

I could put it in a separate (but still public) section of the class (e.g. at the end) to make it clear it is not part of the public interface.  Another idea is to name it specially (e.g. leading or trailing underscore) to make it clear it is  not really part of the public interface.

I think the best solution is to have the RobotInteraction be a parameter of the InteractionHandler constructor.  This function will not be needed anymore.  But legacy code may still use the old constructor so I don't know if I can get rid of it (your opinion?), which means this function is still needed.

I struggled with this for way longer than such a trivial detail merits, so if you have any good solutions to this problem in general I'd love to hear them.

isucan, line 244: I see. I think this is a class that is very unlikely to be present in user
code, and i think the constructor argument is the best option (slightly
breaking api). Better than using friend, I think.

robot_interaction/include/moveit/robot_interaction/interaction.h
acornacorn, line 99: Good point -- will add comments

robot_interaction/include/moveit/robot_interaction/interaction.h
acornacorn: This is the existing prototype.  I agree 

```
const visualization_msgs::InteractiveMarkerFeedback &
```

would be better, but changing it would break legacy code.

I plan to change this later by replacing these callback functions with virtual methods in the GenericInteraction class, so this will be improved later.

robot_interaction/include/moveit/robot_interaction/interaction.h
acornacorn: will fix.  Eventually I think these may become classes but for now you are right.

robot_interaction/include/moveit/robot_interaction/interaction.h
acornacorn, line 172: Probably eventually.  This is how it was and I'm inclined to leave it as it until the next phase of changes.

robot_interaction/include/moveit/robot_interaction/interaction.h
acornacorn, line 193: As above, I'm inclined to leave as is for now.  Will fix later.

robot_interaction/include/moveit/robot_interaction/interaction_handler.h
acornacorn: will fix

robot_interaction/include/moveit/robot_interaction/kinematic_options.h
acornacorn: Since this struct has methods I wanted to be clear that its all public (since some might see the methods and assume it is a class.  Is that bad?  I can remove it if it bugs you.

robot_interaction/include/moveit/robot_interaction/kinematic_options.h
acornacorn: This is intended to be called through a boost::bind.  Using boost::bind with references requires boost::ref() or boost::cref() to pass by value.  For const parameters forgetting boost::cref() just means inefficiency.  For non-const parameters forgetting boost::ref() would mean incorrect behavior with no compiler warning.  I prefer to leave these all as pointers for this reason.

The return value I could be convinced to change.  But this function is intended to be callable through the LockedRobotState::modifyState() function, and that takes a function object with void return.  I could make that function return a bool, but it was unclear to me what a bool return value should mean.  Does it mean the function failed to set the state?  Or (as when using setStateFromIK) does it mean it succeeded in setting the state but the state is somehow invalid?  

Do you think the LockedRobotState::modifyState() method should have a return value?  If so,  what comment would you use to describe the meaning of the return value (it would be the same as the return value of RobotInteraction::updateState() and GenericInteraction's ProcessFeedbackFn callback -- what comment would you use to describe those return values?)

robot_interaction/include/moveit/robot_interaction/kinematic_options_map.h
acornacorn: will fix.

robot_interaction/include/moveit/robot_interaction/kinematic_options_map.h
acornacorn: See reply to other comment

robot_interaction/include/moveit/robot_interaction/locked_robot_state.h
acornacorn, line 62: Good point.  Will fix (just keep latter 2)

robot_interaction/include/moveit/robot_interaction/locked_robot_state.h
acornacorn: fixed

robot_interaction/include/moveit/robot_interaction/locked_robot_state.h
acornacorn, line 103: I strongly disagree.  The whole purpose of this class is to ensure that state_ is not modified unless (A) it is uniqe and (B) it is done with state_lock_ held.  Making state_ protected will allow violation of this too easily.  If you want to modify the state from the parent class use modifyState().

robot_interaction/include/moveit/robot_interaction/robot_interaction.h
acornacorn, line 165: Since the lifetime of RobotInteraction() is ambiguous it is not clear to me that doing so is safe.  Consider

```
lock_access_to robot_interaction
const KinematicOptionsMapPtr& kinematic_options_map = getKinematicOptionsMap();
unlock_access_to robot_interaction
// some other thread deletes robot_interaction
kinematic_options_map->dereference
```

Am I being too paranoid?  Tell me if this is silly and I'll fix it...

robot_interaction/include/moveit/robot_interaction/robot_interaction.h
acornacorn: oops -- yes, will fix

robot_interaction/src/interaction_handler.cpp
acornacorn: oops.  will fix

robot_interaction/src/interaction_handler.cpp
acornacorn, line 123: I think it makes sense as a private static method, but I can change it if you really prefer that.  Just curious why that is better?

robot_interaction/src/interaction_handler.cpp
acornacorn, line 361: KinematicOptionsMap locks its members internally and only returns copies of its internal state.  Otherwise the KinematicOptionsMap lock (currently private) would have to be locked (and unlocked) by any function calling KinematicOptionsMap methods.  Using the state_lock_ to protect the contents of kinematic_options_map_ would not work because it is shared with RobotInteraction and all the other InteractionHandlers (and state_lock_ is specific to this InteractionHandler instance).

robot_interaction/src/interaction_handler.cpp
acornacorn, line 546: Is there a reason swapping boost::shared_ptr is better than assigning?

robot_interaction/src/kinematic_options.cpp
acornacorn: good point.  Will fix

robot_interaction/src/kinematic_options.cpp
acornacorn, line 83: thanks :)

robot_interaction/src/kinematic_options_map.cpp
acornacorn: will fix

robot_interaction/src/robot_interaction.cpp
acornacorn, line 192: :)

robot_interaction/include/moveit/robot_interaction/interaction_handler.h
acornacorn, line 244: Sounds good.  I'll fix this in stages.  I've already added the new constructor.  Next (after this is merged) I'll update the places that use this code to use the new constructors.  Then I'll eliminate the old constructors and this function.

robot_interaction/include/moveit/robot_interaction/interaction.h
isucan: That sounds fine. Maybe add a todo? This seems fairly low impact though.. some small things we could maybe break, if they are low level enough that they are unlikely to have been used and it is easy to port code forward.

robot_interaction/include/moveit/robot_interaction/kinematic_options.h
isucan: I see your point. I have a mild preference for removing it, but it is your call.

robot_interaction/include/moveit/robot_interaction/kinematic_options.h
isucan: I see your point. If that is the intention though, why not make the function protected or private? I strongly agree with the cref / ref business. We really should avoid needing that. I don't think modifyState needs to return a value, but if we can generically define a notion of success vs failure, returning bool would be nice.

robot_interaction/include/moveit/robot_interaction/kinematic_options.h
isucan: Or maybe we can simply make this a class?

robot_interaction/include/moveit/robot_interaction/locked_robot_state.h
isucan, line 103: Ok, I buy your argument, but it strikes me as odd having a variable and the mutex that protects it at different visibility levels. Can we make that private as well?

robot_interaction/include/moveit/robot_interaction/robot_interaction.h
isucan, line 165: Technically, you are correct, this can happen. If the user does this though, it is unlikely changing the options map makes any sense. I would allow this to happen -- fairly easy to track down, if it does happen. Your call.

robot_interaction/src/interaction_handler.cpp
isucan, line 123: I prefer keeping things at the minimum visibility necessary. If someone looks at the header file, they need not see this function, although static private is fine.

robot_interaction/src/interaction_handler.cpp
isucan, line 546: Actually, I am not sure. I have not thought about this in detail, but theoretically you do not need to increase count / decrease count for either of them -- you just swap content. I think swap() may be slightly faster, but I am not sure.

acornacorn:
I've addressed Ioan's comments (all the ones I marked "will fix")

robot_interaction/include/moveit/robot_interaction/kinematic_options.h
acornacorn: Fixed:
I removed the public:

robot_interaction/include/moveit/robot_interaction/kinematic_options.h
acornacorn: I can add a bool return value.  I just don't understand what it is supposed to mean.  It doesn't mean IK failed to find a solution.  The RobotState may still be changed even if it returns false (right?).  I guess my real problem is that I do not understand the meaning of the value returned from RobotState::setFromIK().

The whole purpose of this class is to call this function.  Making it private would defeat the purpose (every user of the struct would have to be a friend).

However, I think I actually do not call this from modifyState -- there is an intermediate function -- so I can probably change this to use references.  Will look at that...

robot_interaction/include/moveit/robot_interaction/locked_robot_state.h
acornacorn, line 103: We could. But it is useful to share that lock with the subclass so it can use it to lock other members (as mentioned in the comment).  The InteractionHandler (which inherits from this class) uses state_lock_ to protect several of its members.  If InteractionHandler needed its own separate mutex for that it would just mean we have to lock 2 mutexes many places where it is used.

One option is to make state_lock_ private and have a protected member function that grants access to it

```
    boost::mutex getStateLock() const { return state_lock_; }
```

but this seems gratuitous.  It does not really add any protection.

robot_interaction/include/moveit/robot_interaction/robot_interaction.h
acornacorn, line 165: I think you are right that this is unlikely to be a problem.  However I think it makes sense to leave this as is.  This function will really only be used when the options are being modified.  This typically happens from the UI and is not performance critical.  Also all accesses to KinematicOptionsMap require an internal mutex lock and typically an access to a map which mean that the reference inc/dec will not significantly slow things down.

On the other hand if this fails (RobotInteraction is deleted while UI is modifying the value from a different thread)  it will result in a rare race-condition based memory corruption which will be almost impossible to find.

Note: This function is likely to change in my next batch of fixes where I address thread safety in the RobotInteraction class.  We can revisit then, but to be thread safe I think this function will need to return the pointer and not a reference to the pointer.

isucan, line 165: Sounds good.

robot_interaction/include/moveit/robot_interaction/locked_robot_state.h
isucan, line 103: Ok, I am convinced the current way of doing things makes sense. Thanks for the explanations!

robot_interaction/include/moveit/robot_interaction/kinematic_options.h
isucan: setFromIK should return true if the function succeeded if the IK call succeeded and we set the state to a valid state, according to the constraints we use (which could be empty). There is a bit of weirdness here in the sense that the state could be destroyed even if we return false. I agree with your points above, but if we can keep the ptr stuff a little more hidden for the stuff that can be const&, I'd be happy.

robot_interaction/src/kinematic_options.cpp
davetcoleman, line 83: +1 for using "coolness" as a vote

acornacorn:
I've fixed the setStateFromIK() prototypes (use references instead of pointers as requested by Ioan).
Also updated some comments and formatting.

I believe all Ioan's suggestions have now been addressed.

PR #424, opened 2014-02-23
adding error code returns to relevant functions

sachinchitta:
Only for cpp - will do python later.

### DIFF
--- a/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group.h
+++ b/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group.h
@@ -1,6 +1,7 @@
 /*********************************************************************
  * Software License Agreement (BSD License)
  *
+ *  Copyright (c) 2014, SRI International
  *  Copyright (c) 2012, Willow Garage, Inc.
  *  All rights reserved.
  *
@@ -32,7 +33,7 @@
  *  POSSIBILITY OF SUCH DAMAGE.
  *********************************************************************/
 
-/* Author: Ioan Sucan */
+/* Author: Ioan Sucan, Sachin Chitta */
 
 #ifndef MOVEIT_MOVE_GROUP_INTERFACE_MOVE_GROUP_
 #define MOVEIT_MOVE_GROUP_INTERFACE_MOVE_GROUP_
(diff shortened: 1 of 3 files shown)


isucan:
I think adding the error codes is a great idea! A thought on the api though: the bool value the functions return is basically a very rough approximation for an error code. The api changes in this PR are adding a mandatory argument and the bool value is duplicate info. I think we should revisit the api changes before the merge: i think we should return an error code instead of a bool and in most cases we can then leave the argument list as it was, reducing api breakage.

adolfo-rt:
+1 for this one indeed!. API compatibility aside, I agree with @isucan that once error codes are exposed, the boolean return value becomes redundant. For the simpler usecases where one only wants to know if the call succeeded or not, it makes the code a bit more verbose, as one would have to explicitly compare against [SUCCESS](https://github.com/ros-planning/moveit_msgs/blob/hydro-devel/msg/MoveItErrorCodes.msg#L4). I can happily live with that.

The only method where the error code can't be used as return value is [computeCartesianPath](https://github.com/ros-planning/moveit_ros/pull/424/files#diff-d89c4ff045d2d131e4a1089bc11c27afR605).

The Indigo release could be a good milestone to push this API change forward. We'd probably backport it to Hydro, as it's relevant enough.

Refs #420.

isucan:
I think we can try to push this in hydro still, with minimal breakage. I agree with @adolfo-rt that computeCartesianPath will need an API change. For the rest of the functions, what I would do is something along the lines of this:

```
class MoveItErrorCode : public moveit_msgs::MoveItErrorCodes {
public:
  operator bool() const { return val == MoveItErrorCodes::SUCCESS; }
}
```

and we always return MoveItErrorCode. If the use gets this as a moveit_msgs instance, we get type slicing, but that is ok. There will be no mem loss.
If they compare as bool, we have the automatic cast to bool, so old code will work as expected, without any change.

sachinchitta:
I have an implementation of @isucan in a branch https://github.com/ros-planning/moveit_ros/tree/error_codes_minimal but I prefer the one in this branch - I think its easier for naive users looking through the API to understand this one.

adolfo-rt:
I like @isucan's suggestion ([error_codes_minimal](https://github.com/ros-planning/moveit_ros/tree/error_codes_minimal) branch implementation) more. I find it confusing to have two variables with different levels of detail to indicate the success of an operation. Further, the error code is a non-optional parameter, so naive users will still have to pass it to the method calls, no way to ignore it.

I think that good code documentation, and especially covering this issue in the basic tutorials should be enough to send the message that it's OK to do both:

``` c++
bool move_ok = move_group.move();
```

and

``` c++
MoveItErrorCode error_code = move_group.move();
```

sachinchitta:
Closed - see #429

Issue #421, opened 2014-02-13
Allow namespacing moveit::planning_interface::MoveGroup instances


adolfo-rt:
Currently, there is no way to connect multiple `moveit::planning_interface::MoveGroup` instances to different move_group servers from within the same ros node. This is because the [action clients](https://github.com/ros-planning/moveit_ros/blob/hydro-devel/planning_interface/move_group_interface/src/move_group.cpp#L126) are created without specifying the NodeHandle on top of which we want to namespace them. A usecase where this would be needed would be that of sending planning requests to different robots, each living in its own namespace. Note that it _is_ currently possible to create multiple `moveit::planning_interface::MoveGroup` instances, each associated to a different planning group of the _same_ move_group server.

The above argument applies for the rest of the ROS API ([services](https://github.com/ros-planning/moveit_ros/blob/hydro-devel/planning_interface/move_group_interface/src/move_group.cpp#L135-L137), [publishers](https://github.com/ros-planning/moveit_ros/blob/hydro-devel/planning_interface/move_group_interface/src/move_group.cpp#L121-L122)), which use the default-constructed [node_handle_](https://github.com/ros-planning/moveit_ros/blob/hydro-devel/planning_interface/move_group_interface/src/move_group.cpp#L956).

I suggest adding an additional NodeHandle member to the [MoveGroup::Options](https://github.com/ros-planning/moveit_ros/blob/hydro-devel/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group.h#L67) struct, and assign it to the [node_handle_](https://github.com/ros-planning/moveit_ros/blob/hydro-devel/planning_interface/move_group_interface/src/move_group.cpp#L956) member.

jbohren:
This is complete now, right?

PR #437, opened 2014-03-11
Use member NodeHandle in action clients.

adolfo-rt:
Currently services and topics are already using the member NodeHandle instance,
but not the action clients.

This is relevant for two reasons:
- Consistency in the resulting ROS API namespace (everything in the same namespace).
- Consistency in the spinning policy. All services, topics and actions will be spinned
  by the same NodeHandle, and whatever custom (or not) spinners and callback queues it
  has associated.

### DIFF
--- a/planning_interface/move_group_interface/src/move_group.cpp
+++ b/planning_interface/move_group_interface/src/move_group.cpp
@@ -83,7 +83,10 @@ class MoveGroup::MoveGroupImpl
 {
 public:
 
-  MoveGroupImpl(const Options &opt, const boost::shared_ptr<tf::Transformer> &tf, const ros::Duration &wait_for_server) : opt_(opt), tf_(tf), node_handle_(opt.node_handle_)
+  MoveGroupImpl(const Options &opt, const boost::shared_ptr<tf::Transformer> &tf, const ros::Duration &wait_for_server)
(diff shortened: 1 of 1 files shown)


adolfo-rt:
I just realized that the fix for #421 was missing an important bit that didn't make it in #428.

Usecase: Isolate the servicing of `move_group` callbacks from the rest of your node. I'm constructing `move_group` objects passing a NodeHandle bound to a specific AsyncSpinner+CallbackQueue combination. This allows to preserve a simple synchronous spinning model in a node, while providing  `move_group` with its asynchronous spinning requirements.

It's bit verbose, but looks something like this

``` c++
// Associate a callback queue to a node handle
ros::NodeHandle nh;
ros::CallbackQueue cq;
nh.setCallbackQueue(&cq);

// Make an async spinner service the callback queue
ros::AsyncSpinner as(1, &cq);
as.start();

// Setup a move group instance
using moveit::planning_interface::MoveGroup;
MoveGroup::Options opts("group_name");
opts.node_handle_ = nh;
MoveGroup move_group(opts);
```

isucan:
+1

adolfo-rt:
Friendly ping.

adolfo-rt:
:+1:

PR #438, opened 2014-03-12
Add functions to python bindings

e1000i:
Added python bindings to manage path constraints

this function was not accessible from python code:

```
void moveit::planning_interface::MoveGroup::setPathConstraints(const moveit_msgs::Constraint &constraint)
```

I added the necesary code to generate the python bindings.

At the same time created a function to retrieve this paths constraints
and generated also the python bindings.

```
moveit_msgs::Constraint moveit::planning_interface::MoveGroup::getPathConstraints() const
```

I opened this issue in the moveit_commander in order to track this missing functionality: 
https://github.com/ros-planning/moveit_commander/issues/18


sachinchitta:
Thanks!

PR #453, opened 2014-05-07
trajectory_execution_manager now removes virtual joints.

hersh:
There are no controllers for virtual joints, so strip them out of the trajectories sent off for execution.

### DIFF
--- a/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
+++ b/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
@@ -744,9 +744,23 @@ bool TrajectoryExecutionManager::distributeTrajectory(const moveit_msgs::RobotTr
   parts.clear();
   parts.resize(controllers.size());
 
+  // Virtual joints by definition do not really exist, so don't try to
+  // control them even if the trajectory has entries for them.
+  std::set<std::string> virtual_joints;
+  const std::vector<srdf::Model::VirtualJoint> &vjoints = robot_model_->getSRDF()->getVirtualJoints();
+  for(size_t i = 0; i < vjoints.size(); i++)
+  {
+    virtual_joints.insert(vjoints[i].name_);
+  }
+
   std::set<std::string> actuated_joints_mdof;
   actuated_joints_mdof.insert(trajectory.multi_dof_joint_trajectory.joint_names.begin(),
                               trajectory.multi_dof_joint_trajectory.joint_names.end());
+  std::set<std::string> difference;
+  std::set_difference(actuated_joints_mdof.begin(), actuated_joints_mdof.end(),
+                      virtual_joints.begin(), virtual_joints.end(),
+                      std::inserter(difference, difference.end()));
+  actuated_joints_mdof = difference;
   std::set<std::string> actuated_joints_single;
   for (std::size_t i = 0 ; i < trajectory.joint_trajectory.joint_names.size() ; ++i)
   {
@@ -758,6 +772,12 @@ bool TrajectoryExecutionManager::distributeTrajectory(const moveit_msgs::RobotTr
       actuated_joints_single.insert(jm->getName());
     }
   }
+
+  difference.clear();
+  std::set_difference(actuated_joints_single.begin(), actuated_joints_single.end(),
+                      virtual_joints.begin(), virtual_joints.end(),
+                      std::inserter(difference, difference.end()));
+  actuated_joints_single = difference;
 
   for (std::size_t i = 0 ; i < controllers.size() ; ++i)
   {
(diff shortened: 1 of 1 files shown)


planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -744,9 +744,23 @@ bool TrajectoryExecutionManager::distributeTrajectory(const moveit_msgs::RobotTr
...
+  const std::vector<srdf::Model::VirtualJoint> &vjoints = robot_model_->getSRDF()->getVirtualJoints();
+  for(size_t i = 0; i < vjoints.size(); i++)
+  {
+    virtual_joints.insert(vjoints[i].name_);
+  }
+
   std::set<std::string> actuated_joints_mdof;
   actuated_joints_mdof.insert(trajectory.multi_dof_joint_trajectory.joint_names.begin(),
                               trajectory.multi_dof_joint_trajectory.joint_names.end());
+  std::set<std::string> difference;
```
adolfo-rt, line 759: How about adding a few comments in this hunk and the next one saying what is being done: removing virtual joints from the multi_dof / joint trajectories.

This is a style thing, but using the difference vector twice was a bit confusing when I did the first pass on the code. IMO scoping the two blocks for multi-dof and joint trajectories inside separate {} blocks and having two `difference` instances that live within that scope would be clearer. No need to clear() in between.

adolfo-rt:
Very minor comment posted inline, otherwise +1. This is a very relevant changeset.

isucan:
Hmm. I am not sure about this change. I am assuming you are trying to solve the problem of base x/y/theta not being controlled? If so, this is not the way. We could (probably do) have controllers for virtual joints and the trajectory manager should deal with that.

hersh:
@isucan I am working with a humanoid, and this is my attempt to work around the notion of a "root joint" with the world which MoveIt requires.

I wrote an IK solver which takes as a configuration parameter which link to consider "fixed" for a given IK call.  For me this is generally one of the feet, but it is never MoveIt's idea of the "root" link or joint.  It is possible to rework the URDF so that one of the feet has the root joint, but then if we tried to use IK for walking motions, we would have to somehow change to a different URDF with the root on the other foot whenever we wanted to move the original root foot.

So in addition to computing angles for all the real joints, my IK solver computes variable values for the 6 variables of the "floating" virtual joint between the hips and the world.  This way when I display trajectories in RViz, the robot's feet stay on the ground even when they move relative to the hips.

So I'm left with this situation where I want the virtual joint variables to be present in the trajectory everywhere until it gets sent to be executed.  Which is why I'm trying to drop them here in the trajectory execution manager.

I would think if one were controlling the base x/y/theta of a PR2 for example, it would make more sense to label the base joint as an actual planar joint and not as a virtual one, no?  I mean, the wheels and their motors are real, they are not virtual.

The other way to solve this, I realize, is to write a "no-op" controller for the root joint and add that to my controller setup.  It just seemed like more work and kind of silly, since I had the impression that virtual joints never corresponded to anything physical.

hersh:
See also https://github.com/ros-planning/moveit_core/pull/170 where I did the same thing in the time parameterizer.

planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
hersh, line 759: I'm going to wait and see how the conversation with Ioan goes before putting more time into this change, but I certainly see your point.  I'll comment the code more if this design is OK.

isucan:
Sorry for the delayed reply. Virtual joints are any joints defined in the srdf. So Planar or floating joints are also virtual. If no connection to the world is specified, then fixed is used. I understand why this patch fixes the problem you are having, but I think having a pass-through controller that removes that joint is the better approach. One thing that should not happen though is commands sent for fixed joints. Do you get the trajectory generator complaining about missing joints / not finding groups?

hersh:
I get the error "Unable to identify any set of controllers that can actuate the specified joints" from the file changed in this pull request.  It knows it has no controller for the world_joint, and so refuses to continue.

isucan:
How about defining a dummy controller that does nothing for that joint
only? The traj exec manager will split that part of the trajectory for the
no op controller.

hersh:
That's what I meant above when I said I could write a "no-op" controller.  Now I need to figure that out.

So what is the advantage of representing the robot's position in the world as a joint?  I always thought that was a kind of awkward aspect of moveit, and I feel like now I am running into that awkwardness.

isucan:
The advantage is we do not need to handle weird cases in planning -- a joint is a DOF. How the robot moves in the world is also a number of DOF. Treating these the same way makes things cleaner for the planner. Defining a no-op controller is really easy. You can simply copy the fake controller and remove all the code from the functions (just return success for functions that expect a bool return).  There is also a tutorial on the wiki, but it seems to have not gotten ported to the new docs. (and I never know how to find the old wiki; off-topic: maybe we can add a link to the old page somewhere?)

hersh:
After Ioan's comments, I agree that this pull request is not a good way to solve my problem.

Turns out a no-op FollowJointTrajectory controller is really easy to write - it's just an actionlib.SimpleActionServer for the FollowJointTrajectory action that always succeeds immediately.  I wrote one in Python in a few minutes and it is now doing the work that this pull request did.

So, closing.

davetcoleman:
A housekeeping aside - I've noticed a lot of branches on `moveit_*` being merged or closed without being deleted, like this one. I removed a huge number of abandoned branches a few months ago and think it would make this easier for newcomers if we kept the branch list minimum. Any need to keep this branch around?

hersh:
It's always fine to delete the branch for a pull request that's been merged.

If it's been closed without merging, just look at the text.  For one like this that says definitively that it is no longer wanted, you can definitely delete the branch.

If it has any question in the thread about future usefulness, please leave the branch.

However, why does a newcomer need to know anything about the branches?  The build instructions say to use a .rosinstall file which specifies the branch to use for each repo.

davetcoleman:
That's a good point, but by newcomer what I meant was someone newer to developing/contributing to moveit's code.

PR #457, opened 2014-05-15
Added field next to planning_time for planning_attempts

drchrislewis:
New field is tied to move_group::setNumPlanningAttempts().
Now, ParallelPlanner terminates either due to timeout, or due to this many attempts.
Note, that ParallelPlanner run's Dijkstra's on all the nodes of all the successful plans (hybridize==true).


isucan:
+1

PR #459, opened 2014-05-20
Added more informative error message text when can't find controllers.

hersh:
This change helped me find why my controller setup was not working, so I thought I would share it.  Has no effect when things are working.


mikeferguson:
+1

PR #462, opened 2014-05-22
fixes #461 and a potential segfault

### LINKED ISSUE #461: Pick action does not fill in "grasp" field of return message
Should be filling in the field with the grasp actually used.

mikeferguson:
Also includes a couple of style fixes (trailing newlines).


isucan:
+1

PR #473, opened 2014-06-25
Removed PlanningContext clear before planning call

arjungm:
Originally, before each benchmark was tested, PlanningContext was cleared prior to calling solve.

Deleting the line lets the test continue as normal and produce results.

Credit for the fix: https://groups.google.com/d/msg/moveit-users/DjI9yE_BoHU/oa9IvVrUTmoJ


sachinchitta:
LGTM.

PR #491, opened 2014-09-10
Fix bug in place-planning where attached object was not considered in plan.

hersh:
This fixes a bug in pick and place where the attached object was not considered when planning the "place" phase.


mikeferguson:
I've been bitten by an issue like this -- and never managed to track down what the cause was... +1

PR #492, opened 2014-09-10
Fix bug in place-planning where attached object was not considered in plan.

hersh:
Hydro version of #491.


sachinchitta:
This needs to be against hydro-devel.

Issue #503, opened 2014-10-16
Rviz crashes when adding MotionPlanning to display


jasedit:
I'm running the latest checkout of the repository on Ubuntu 14.04, 64-bit with Indigo.

If I launch rviz and attempt to add MotionPlanning to the display, it immediately crashes with the error `Attempt to add global initializer failed, status: DuplicateKey throwSockExcep`

Running in gdb produces the following backtrace:

```
Program received signal SIGABRT, Aborted.
0x00007ffff623abb9 in __GI_raise (sig=sig@entry=6)
    at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
56  ../nptl/sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  0x00007ffff623abb9 in __GI_raise (sig=sig@entry=6)
    at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007ffff623dfc8 in __GI_abort () at abort.c:89
#2  0x00007fffa4eb12a7 in mongo::GlobalInitializerRegisterer::GlobalInitializerRegisterer(std::string const&, boost::function<mongo::Status (mongo::InitializerContext*)> const&, std::vector<std::string, std::allocator<std::string> > const&, std::vector<std::string, std::allocator<std::string> > const&) ()
   from /home/jpz/catkin_ws//devel/lib/libmoveit_move_group_interface.so
#3  0x00007fffa4dcaa38 in _GLOBAL__sub_I_sock.cpp ()
   from /home/jpz/catkin_ws/devel/lib/libmoveit_move_group_interface.so
#4  0x00007ffff7dea13a in call_init (l=<optimized out>, argc=argc@entry=3, 
    argv=argv@entry=0x7fffffffcd38, env=env@entry=0x7fffffffcd58)
    at dl-init.c:78
#5  0x00007ffff7dea223 in call_init (env=<optimized out>, 
    argv=<optimized out>, argc=<optimized out>, l=<optimized out>)
    at dl-init.c:36
#6  _dl_init (main_map=main_map@entry=0xebe250, argc=3, argv=0x7fffffffcd38, 
    env=0x7fffffffcd58) at dl-init.c:126
#7  0x00007ffff7deec70 in dl_open_worker (a=a@entry=0x7fffffffb148)
    at dl-open.c:577
#8  0x00007ffff7de9ff4 in _dl_catch_error (
[...]

PR #504, opened 2014-10-16
Fixing segfault if filtered_cloud_topic is not set in PointCloudOctomapUpdater

jasedit:
Iterators were created into a PointCloud2 message which was only created if filtered_cloud_topic was set. Furthermore, the incoming data was never used during the update - the octomap update iterated over the output PointCloud2, as opposed to the input PointCloud 2. On my setup (Trusty/Indigo), PointCloudOctomapUpdater would crash immediately unless I defined filtered_cloud_topic, and I would never see any output data. I believe this pull request fixes this issue.

### DIFF
--- a/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
+++ b/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
@@ -206,17 +206,26 @@ void PointCloudOctomapUpdater::cloudMsgCallback(const sensor_msgs::PointCloud2::
   octomap::KeySet free_cells, occupied_cells, model_cells, clip_cells;
   boost::scoped_ptr<sensor_msgs::PointCloud2> filtered_cloud;
 
+  //We only use these iterators if we are creating a filtered_cloud for
+  //publishing. We cannot default construct these, so we use scoped_ptr's
+  //to defer construction
+  boost::scoped_ptr<sensor_msgs::PointCloud2Iterator<float> > iter_filtered_x;
+  boost::scoped_ptr<sensor_msgs::PointCloud2Iterator<float> > iter_filtered_y;
+  boost::scoped_ptr<sensor_msgs::PointCloud2Iterator<float> > iter_filtered_z;
+
   if (!filtered_cloud_topic_.empty()) {
     filtered_cloud.reset(new sensor_msgs::PointCloud2());
     filtered_cloud->header = cloud_msg->header;
     sensor_msgs::PointCloud2Modifier pcd_modifier(*filtered_cloud);
     pcd_modifier.setPointCloud2FieldsByString(1, "xyz");
     pcd_modifier.resize(cloud_msg->width * cloud_msg->height);
+
+    //we have created a filtered_out, so we can create the iterators now
+    iter_filtered_x.reset(new sensor_msgs::PointCloud2Iterator<float>(*filtered_cloud, "x"));
+    iter_filtered_y.reset(new sensor_msgs::PointCloud2Iterator<float>(*filtered_cloud, "y"));
+    iter_filtered_z.reset(new sensor_msgs::PointCloud2Iterator<float>(*filtered_cloud, "z"));
   }
   size_t filtered_cloud_size = 0;
-  sensor_msgs::PointCloud2Iterator<float> iter_filtered_x(*filtered_cloud, "x");
-  sensor_msgs::PointCloud2Iterator<float> iter_filtered_y(*filtered_cloud, "y");
-  sensor_msgs::PointCloud2Iterator<float> iter_filtered_z(*filtered_cloud, "z");
 
   tree_->lockRead();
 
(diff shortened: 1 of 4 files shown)


jasedit:
Odd - the build error is that perception/point_containment_filter/src/shape_mask.cpp could not find sensor_msgs/point_cloud2_iterator.h, which are not files I've touched in any way. It looks like CMakeLists.txt and packages.xml did not declare any dependency on sensor_msgs, which I've added. I think that should fix the Travis build error.

mikeferguson:
@jasedit Disregard build errors -- Travis doesn't support Trusty, so this is still being built against precise/hydro, and the iterator does not exist in hydro.

mikeferguson:
Looping in @vrabaud since he created the point_cloud2_iterator.

davetcoleman:
I've just run into this issue, I'll test your PR

Also, I've just disabled Travis for all MoveIt! repos per the mailing list email I sent out.

jasedit:
Thanks! I've verified that it no longer crashes on my system. However, I can't validate any of the data coming out of the updater due to #503, which I haven't made any headway on solving.

davetcoleman:
I'm still getting a segfault even with your branch when I start move_group with a point cloud being published:

> 0x00007fffd1abd503 in occupancy_map_monitor::PointCloudOctomapUpdater::cloudMsgCallback (this=0xa25f40, cloud_msg=...) at /home/dave/ros/ws_moveit/src/moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp:249

I'm currently debugging but any hints are welcomed

davetcoleman:
Also, it is not clear to me why you made the changes you made in this PR. Perhaps adding inline comments to your changes would help me and others in the future?

jasedit:
Line 249 would imply that cloud_msg is an empty/invalid message. There doesn't appear to be a check for this, and may be a reason why I haven't seen it yet.

jasedit:
I can add comments on the iterators momentarily.

jasedit:
Comments added - if there is additional info necessary, let me know and I'll be happy to flesh it out.

davetcoleman:
Thanks for the comments, I read a lot into the code just now. I believe your fixes are good but I think there are some more issues, in fact. See https://github.com/ros-planning/moveit_ros/issues/505.

In your testing are you actually filtering a point cloud, or do you have that feature disabled?

jasedit:
I am filtering a point cloud using this block of code. I don't think you can cause the segfault without filtering the points.

perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
```diff
@@ -224,46 +233,45 @@ void PointCloudOctomapUpdater::cloudMsgCallback(const sensor_msgs::PointCloud2::
...
-      sensor_msgs::PointCloud2Iterator<float> iter_z(*filtered_cloud, "z");
-      iter_x += row*cloud_msg->row_step;
-      iter_y += row*cloud_msg->row_step;
-      iter_z += row*cloud_msg->row_step;
-
-      for (unsigned int col = 0; col < cloud_msg->width; col += point_subsample_, iter_x += point_subsample_,
-        iter_y += point_subsample_, iter_z += point_subsample_)
+      sensor_msgs::PointCloud2ConstIterator<float> pt_iter(*cloud_msg, "x");
+      //set iterator to point at start of the current row
+      pt_iter += row;
```
vrabaud: that is wrong, you should have 

```
t_iter += row*cloud_msg->width;
```

jasedit: Or `pt_iter += row_c`, no?

vrabaud: that works too.

jasedit:
Commit is fixed to correctly address #505, I believe.

mikeferguson:
Ok, here's my thoughts:
- +1 to all changes in pointcloud_octomap_updater.cpp
- +1 to adding depend on sensor_msgs
- -0 to adding depends in cmake on sensor_msgs_generate_messages_cpp, there are a lot of other things that aren't declared, and most people aren't building sensor_msgs locally

I would recommend squashing this down to a single commit also.

jasedit:
Commits have been squashed, sensor_msgs_generate_messages_cpp has been removed.

perception/point_containment_filter/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ add_library(${MOVEIT_LIB_NAME} src/shape_mask.cpp)
 set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES COMPILE_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
 set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES LINK_FLAGS "${OpenMP_CXX_FLAGS}")
 target_link_libraries(${MOVEIT_LIB_NAME} ${catkin_LIBRARIES} ${Boost_LIBRARIES})
+add_dependencies(${MOVEIT_LIB_NAME} sensor_msgs_generate_messages_cpp)
```
davetcoleman: You said you took this out, right?

perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
```diff
@@ -224,46 +233,45 @@ void PointCloudOctomapUpdater::cloudMsgCallback(const sensor_msgs::PointCloud2::
   {
     /* do ray tracing to find which cells this point cloud indicates should be free, and which it indicates
      * should be occupied */
-    for (unsigned int row = 0; row < cloud_msg->height; row += point_subsample_)
+    for(unsigned int row = 0; row < cloud_msg->height; row += point_subsample_)
```
davetcoleman: This is super minor, but a digression from this file's standard

davetcoleman:
It works - thanks so much! I'll merge as soon as my two comments are addressed.

jasedit:
Addressed!

jasedit:
Thanks for helping get to the bottom of this!

PR #518, opened 2014-10-31
Publish rgb filtered cloud

garaemon:
- Add filtered_cloud_use_color parameter
- If the parameter is true, OctomapUpdater publishes colroed pointcloud (`pcl::PointXYZRGB`)
- In order to achieve this, make several method template methods (need to support `pcl::PointXYZ` and pcl::PointXYZRGB)
- This PR includes https://github.com/ros-planning/moveit_ros/pull/517
- Green: input cloud
  a
- RGB: filtered cloud
  b


garaemon:
I found a bug about this code, I close this pull request.

PR #522, opened 2014-11-01
add dependency on std_srvs (part of octomap clearing service)

mikeferguson:
This is intended to fix the build of move_group http://jenkins.ros.org/job/ros-indigo-moveit-ros-move-group_binarydeb_saucy_amd64/67/console


mikeferguson:
Heads up @isucan @davetcoleman -- we'll also need another release (the devel job hasn't run yet from what I can see, we might want to wait for it to complete as it might catch other errors)

mikeferguson:
and looks like all moveit debs are now built in shadow.

PR #525, opened 2014-11-02
GL_TYPE() is a function in newer versions of OpenGL, this fixes tests on...

mikeferguson:
... Ubuntu 14.04. Should take care of #524


mikeferguson:
Hold on merging this for a moment -- I accidentally left some extra includes --- removing now

mikeferguson:
Ok, fixed.

PR #529, opened 2014-11-15
indigo-boost-1.57

v4hn:


mikeferguson:
LGTM, do you have a link/note as to what changed in 1.57 that requires these?

v4hn:
95d1ef5b70a44e32000f64f8f423a239b3b8450c is probably due to this commit
which removes scoped_ptr from boost::thread :
https://github.com/boostorg/thread/commit/ec542d3c8b89bec80d73f83622531e0fbadc248b

I'm not sure about the other one because it looks like it depended on
scoped_ptr being included from deeper within pluginlib, moveit and ros headers.
It's probably due to the same change. ROS includes thread.hpp in a couple of places.

PR #534, opened 2014-11-23
fix duplicate planning attempt box, also fix warning about name

mikeferguson:
An earlier version of the PR seems to have been merged, and thus f26a731a921993acea394ccf09f187fad284ca7b should not have been cherry picked recently into indigo-devel. This PR gets rid of the duplicate planning_attempt box in the GUI. Also makes the name of the horizontal layout unique, thus getting rid of a warning.


gavanderhoorn:
Just wondering, but was this ever backported to Hydro? The planning UI still has two boxes for _Planning Attempts_ in `0.5.20` here.

PR #535, opened 2014-11-23
Fix warnings

mikeferguson:
Some other warnings from -Wall


isucan:
Thanks, this is great!

Issue #532, opened 2014-11-21
Support for alternate collision checkers?


mikeferguson:
I've now written my own collision checker, complete with derived CollisionWorld/CollisionRobot -- but it seems there is no elegant way to load these? My first (hacky) approach was to write a move_group_capability that simply called setActiveCollisionDetector on the planning scene. This managed to get my custom code loaded through the move_group_capability plugin loader and active -- however, it means that the RVIZ plugin is still not using my collision checker with IK.

My current thought is that we need to:
- Define a plugin API for adding collision detectors.
- Define an optional ROS parameter that holds the name of a desired collision detector to load, and then  have the PlanningSceneMonitor load it, and make it active.

@isucan Does this sound about right? Is there a better approach? The code for the RVIZ plugin jumps around a bunch -- but I presume at the core it still uses the PlanningSceneMonitor, and so this approach would get me the desired results in RVIZ?

isucan:
Hey Mike,

In short, you are right, your plan is the right way to go.
At some point we wanted to have a plugin interface for collision checkers, but we never got to it. I was thinking that with collision detection algorithms, we could probbaly figure out which one is the better one and we could kind of hard code the better one. However, that may not be the case.

mikeferguson:
Thanks for responding. I will whip up a first pass and open a PR for discussion.

mikeferguson:
This is now merged.

PR #536, opened 2014-11-25
add plugin interface for collision detectors

mikeferguson:
This is a potential solution to #532. Highlights:
- Given the way the allocator stuff works (templated and CRTP), I chose not to wrap the CollisionRobot/World classes, but rather to create a new class that has an initialize() method in which one calls setActiveCollisionDetector() and does any other setup related to the collision detector.
- Parameter for the name of the collision plugin is held in "collision_detector", and I implemented a parameter search. In addition, if the parameter is not found, we check for "move_group/collision_detector", mainly so that the RVIZ plugin can use the collision detector.

@isucan please review and give feedback.

### DIFF
--- a/planning/CMakeLists.txt
+++ b/planning/CMakeLists.txt
@@ -43,6 +43,7 @@ set(THIS_PACKAGE_INCLUDE_DIRS
     plan_execution/include
     kdl_kinematics_plugin/include
     srv_kinematics_plugin/include
+    collision_plugin_loader/include
 )
 
 catkin_package(
@@ -75,6 +76,7 @@ link_directories(${Boost_LIBRARY_DIRS})
 link_directories(${catkin_LIBRARY_DIRS})
 
 add_subdirectory(rdf_loader)
+add_subdirectory(collision_plugin_loader)
 add_subdirectory(kdl_kinematics_plugin)
 add_subdirectory(srv_kinematics_plugin)
 add_subdirectory(kinematics_plugin_loader)
--- a/planning/collision_plugin_loader/CMakeLists.txt
+++ b/planning/collision_plugin_loader/CMakeLists.txt
@@ -1,0 +1,7 @@
+set(MOVEIT_LIB_NAME moveit_collision_plugin_loader)
+
+add_library(${MOVEIT_LIB_NAME} src/collision_plugin_loader.cpp)
+target_link_libraries(${MOVEIT_LIB_NAME} ${catkin_LIBRARIES})
+
+install(TARGETS ${MOVEIT_LIB_NAME} LIBRARY DESTINATION lib)
+install(DIRECTORY include/ DESTINATION include)
--- a/planning/collision_plugin_loader/include/moveit/collision_plugin_loader/collision_plugin_loader.h
+++ b/planning/collision_plugin_loader/include/moveit/collision_plugin_loader/collision_plugin_loader.h
@@ -1,0 +1,74 @@
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2014 Fetch Robotics Inc.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of Fetch Robotics nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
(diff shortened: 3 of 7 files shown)


planning/collision_plugin/include/moveit/collision_plugin/collision_plugin.h
```diff
@@ -0,0 +1,93 @@
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2014 Fetch Robotics Inc.
```
davetcoleman: Super stealth mode...

planning/collision_plugin/include/moveit/collision_plugin/collision_plugin.h
```diff
@@ -0,0 +1,93 @@
...
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+#ifndef MOVEIT_COLLISION_PLUGIN_H
```
davetcoleman: prefix with package name

mikeferguson: whoops, looks like all the include guards are wrong -- i'll update those tomorrow.

davetcoleman:
+1 very clean code, i didn't compile though and don't have experience with collision checking

planning/collision_plugin/include/moveit/collision_plugin/collision_plugin.h
```diff
@@ -0,0 +1,93 @@
+/*********************************************************************
```
isucan: This lib (collision_plugin) should be in moveit_core though, should it not?

planning/collision_plugin/include/moveit/collision_plugin/collision_plugin_loader.h
```diff
@@ -0,0 +1,73 @@
...
+#include <ros/ros.h>
+#include <moveit/collision_plugin/collision_plugin.h>
+
+namespace collision_detection
+{
+
+/**
+ * @brief This is used to load the collision plugin
+ */
+class CollisionPluginLoader
```
isucan: This should be a separate library, indeed part of moveit_ros/planning

planning/collision_plugin/src/collision_plugin_loader.cpp
```diff
@@ -0,0 +1,150 @@
...
+    if (it->second)
+    {
+      return it->second->initialize(scene, exclusive);
+    }
+    return false;
+  }
+
+private:
+  boost::shared_ptr<pluginlib::ClassLoader<CollisionPlugin> > loader_;
+  std::map<std::string, CollisionPlugin*> plugins_;
```
isucan: These seem to never be freed. We could have the map values be shared pointers / unique pointers maybe?

planning/collision_plugin/include/moveit/collision_plugin/collision_plugin.h
mikeferguson: This header could be in moveit_core -- I don't think you want the loader stuff in there though since it uses pluginlib. I'm fine with putting the header in moveit_core, as that is how the kinematics and planning_request_adapter stuff is done, just let me know and I can update the PR tomorrow morning.

planning/collision_plugin/src/collision_plugin_loader.cpp
```diff
@@ -0,0 +1,150 @@
...
+  ros::NodeHandle& nh,
+  const planning_scene::PlanningScenePtr& scene)
+{
+  if (!scene)
+    return;
+
+  std::string param_name;
+  std::string collision_detector_name;
+
+  if (nh.searchParam("collision_detector", param_name))
```
isucan: Should these params be cached? If this function ends up being called often enough, it would be nicer to avoid calls to the param server.

planning/collision_plugin/src/collision_plugin_loader.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+void CollisionPluginLoader::activate(
+  const std::string& name,
+  const planning_scene::PlanningScenePtr& scene,
+  bool exclusive)
+{
+  loader_->activate(name, scene, exclusive);
+}
+
+void CollisionPluginLoader::setupScene(
```
isucan: Return a bool indicating success?

isucan:
Very minor comments only, design is sound. This looks very close to be merged.

planning/collision_plugin/include/moveit/collision_plugin/collision_plugin.h
isucan: That is what I meant -- made a note about it below.

planning/collision_plugin/include/moveit/collision_plugin/collision_plugin_loader.h
mikeferguson: What do you mean by this? It is built as a separate library, but is not a separate package (as is done for the kinematics_plugin_loader).

isucan: Sorry,  I did not look at the cmake file. My thought was only to have this
and the base class in different packages.

planning/collision_plugin/src/collision_plugin_loader.cpp
```diff
@@ -0,0 +1,150 @@
...
+    return NULL;
+  }
+
+  bool activate(
+    const std::string& name,
+    const planning_scene::PlanningScenePtr& scene,
+    bool exclusive)
+  {
+    std::map<std::string, CollisionPlugin*>::iterator it = plugins_.find(name);
+    if (it == plugins_.end())
```
adolfo-rt: Slightly OT, but I recently learned a [very compact](http://www.cplusplus.com/reference/map/map/count/) way to check whether a map contains a key-value pair. Performance-wise it's the same, but it's much less verbose.

``` c++
if (plugins_.count(name) == 0)
{
  // Key not found
}
else
{
  // Key found. Count == 1, as map is a unique associative container
}
```

planning/collision_plugin/src/collision_plugin_loader.cpp
davetcoleman: +1

planning/collision_plugin/src/collision_plugin_loader.cpp
davetcoleman: cool!

mikeferguson:
Ok, I've moved collision_plugin.h to collision_detection folder of moveit_core. Also moved the collision_plugin folder to a better named collision_plugin_loader within this repo. Fixed potential memory leak by using boost::shared_ptr on the plugin instances.

I didn't cache the rosparams in this commit, since this is only called when a new instance of planningscenemonitor is created, however, that can certainly be a future improvement.

isucan:
Great! Thanks!

PR #540, opened 2014-12-10
planning_scene_monitor: add ros parameter for adding a wait-for-transform lookup time

### LINKED ISSUE #465: "Self-filtering may fail" because of missing waitForTransform
I keep sporadically getting errors like this when using the DepthImageOctomapUpdater:

```
Transform error: Lookup would require extrapolation into the future.  Requested time 1401446451.006236446 but the latest data is at time 1401446450.937553891, when looking up transform from frame [katana_motor2_lift_link] to frame [kinect_rgb_optical_frame]
[ERROR] [1401446451.007936297]: Transform cache was not updated. Self-filtering may fail.
```

This message originates from [depth_image_octomap_updater.cpp:285](https://github.com/ros-planning/moveit_ros/blob/hydro-devel/perception/depth_image_octomap_updater/src/depth_image_octomap_updater.cpp#L285). It looks to me as if there was a `waitForTransform()` missing in [planning_scene_monitor.cpp:818](https://github.com/ros-planning/moveit_ros/blob/56c1b9465fd86795baaff3fb94746e57aeb09ddb/planning/planning_scene_monitor/src/planning_scene_monitor.cpp#L818): When `lookupTransform()` is called with the time stamp from the the DepthImage, sometimes the TF buffer for that time isn't filled yet (in this case, it is 0.07 seconds behind), thereby triggering a race condition.

The Katana arm I'm working with can only publish joint states at 25 Hz, so maybe that's why this race condition happens more frequently than with other robots.

jbohren:
fixes #465

### DIFF
--- a/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
+++ b/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
@@ -490,6 +490,12 @@ private:
   // This field is protected by state_pending_mutex_
   ros::WallDuration dt_state_update_;
 
+  /// the amount of time to wait when looking up transforms
+  // Setting this to a non-zero value resolves issues when the sensor data is
+  // arriving so fast that it is preceding the transform state.
+  // See https://github.com/ros-planning/moveit_ros/issues/465 for more info
+  ros::Duration shape_transform_cache_lookup_wait_time_;
+
   /// timer for state updates.
   // Check if last_state_update_ is true and if so call updateSceneWithCurrentState()
   // Not safe to access from callback functions.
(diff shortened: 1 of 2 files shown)


planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
```diff
@@ -490,6 +490,12 @@ class PlanningSceneMonitor : private boost::noncopyable
   // This field is protected by state_pending_mutex_
   ros::WallDuration dt_state_update_;
 
+  /// the amount of time to wait when looking up transforms
+  // Setting this to a non-zero value resolves issues when the sensor data is
+  // arriving so fast that it is preceding the transform state.
+  // See https://github.com/ros-planning/moveit_ros/issues/465 for more info
```
mikeferguson, line 496: we probably don't need this link, it is likely just adding clutter

jbohren:
This is superseded by #544

PR #547, opened 2014-12-22
Add time factor support for trajectory time parametrization

skohlbr:
This adds support for setting a motion_execution_time_factor in the MotionPlanRequest msg, which can be used to scale the velocity bounds in add_time_parametrization.

Includes necessary changes to
- move_group
- move_group_interface
- Spinbox for motion planning rviz plugin

Requires other pulls requests in moveit_msgs and moveit_core which are stated below.

### DIFF
--- a/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
+++ b/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
@@ -623,7 +623,7 @@
             <item>
              <widget class="QLabel" name="label_11">
               <property name="text">
-               <string>Execution Time Factor:</string>
+               <string>Velocity Scaling:</string>	               
               </property>
              </widget>
             </item>
--- a/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
+++ b/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
@@ -609,10 +609,10 @@
                </sizepolicy>
               </property>
               <property name="maximum">
-               <double>1000.000000000000000</double>
+               <number>1000</number>
               </property>
               <property name="value">
-               <double>10.000000000000000</double>
+               <number>10</number>
               </property>
              </widget>
             </item>
--- a/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
+++ b/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>688</width>
-    <height>377</height>
+    <width>702</width>
+    <height>395</height>
    </rect>
   </property>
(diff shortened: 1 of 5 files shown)


visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
```diff
@@ -609,11 +609,35 @@
                </sizepolicy>
               </property>
               <property name="maximum">
-               <number>1000</number>
+               <double>1000.000000000000000</double>
```
mikeferguson: Should this really have changed? Seems this might be affecting something else?

visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
```diff
@@ -609,11 +609,35 @@
...
               </property>
               <property name="value">
-               <number>10</number>
-               </property>
+               <double>10.000000000000000</double>
```
mikeferguson: Same change here, we should verify these aren't a problem.

visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
skohlbr: I'll revert and test this. It appears a bit weird that the number of planning attempts is set in a QDoubleSpinBox. This should be a plain (integer) SpinBox. Maybe fodder for another (unrelated) Pull Request.

davetcoleman:
Great job with this! I implemented a similar feature just a few weeks ago, but only in moveit_core. This is much needed.

I'm not sure I like the term "motion_execution_time_factor", isn't it more a "motion_execution_velocity_factor" ? In my version I just named it 'velocity_scale'.

I'm also curious what we should do about the max acceleration limits - should it also be scaled? The same amount? This might be naive question, I haven't looked into it much though.

davetcoleman:
Could you add a quick screenshot of the UI change?

mikeferguson:
Hold on merging this until the moveit_msgs release makes it through the pipeline

skohlbr:
@davetcoleman Here's a screenshot of the UI change. I think it makes the motion planning widget a little bit larger.
motion_planning_time_factor

davetcoleman:
That's what I was afraid of... the UI is already too wide IMHO. Can we name it something shorter like Velocity Scale?

skohlbr:
I think it only got higher, not wider. Will make a tad shorter as suggested nevertheless.

skohlbr:
Changed label from "Execution Time Factor:" to "Velocity Scaling:":
motion_planning_velocity_scaling

davetcoleman:
moveit_msgs is in shadow-fixed http://www.ros.org/debbuild/indigo.html?q=moveit_msgs

davetcoleman:
I have confirmed it slows down trajectories when displayed in REALTIME in the motion planning rviz plugin.

mikeferguson:
Thanks for merging -- I'll push new releases shortly

mikeferguson:
Released new moveit_core in order to get this building again.

PR #574, opened 2015-04-09
Visualize convex mesh

dornhege:
Will publish markers with the scaled/padded bodies used internally.
The parameter visualize_scaled_bodies enables this.

### DIFF
--- a/perception/point_containment_filter/include/moveit/point_containment_filter/shape_mask.h
+++ b/perception/point_containment_filter/include/moveit/point_containment_filter/shape_mask.h
@@ -139,7 +139,7 @@ private:
   std::vector<bodies::BoundingSphere> bspheres_;
 
   bool visualize_scaled_bodies_;
-  ros::NodeHandle* param_nh_;
+  ros::NodeHandle param_nh_;
   ros::Publisher pub_vis_;
 };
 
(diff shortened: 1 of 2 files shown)


perception/point_containment_filter/include/moveit/point_containment_filter/shape_mask.h
```diff
@@ -133,6 +137,10 @@ class ShapeMask
   std::set<SeeShape, SortBodies> bodies_;
   std::map<ShapeHandle, std::set<SeeShape, SortBodies>::iterator> used_handles_;
   std::vector<bodies::BoundingSphere> bspheres_;
+
+  bool visualize_scaled_bodies_;
+  ros::NodeHandle* param_nh_;
```
sachinchitta: Don't use a pointer here - you can just instantiate this in the constructor - param_nh_("~").

dornhege:
@sachinchitta: 46d192c should fix these things.

dornhege:
It should apply to indigo-devel cleanly, but I don't have a system to test that on. ros-planning/geometric_shapes#26 would also need to be extended with the changes in ros-planning/geometric_shapes#25.
Someone would need to volunteer to test these.

PR #575, opened 2015-04-13
CurrentStateMonitor no longer requires hearing mimic joint state values.

hersh:
When joint values are updated in a RobotState, it automatically updates mimic joints which depend on them.  So joint values _of_ mimic joints don't need to be listened to by CurrentStateMonitor.

Of course it's not just that they aren't needed, it's that they should not be required, which they previously were before this commit. :)


sachinchitta:
+1

rethink-imcmahon:
+1 as well. Seems logical to remove mimic joints from consideration since the info they provide is redundant.

arunavanag591:
+1

PR #594, opened 2015-07-21
F#276 add db state

dg-shadow:
Adds service interface for access to robots_states in Warehouse. Requires service definitions added to moveit_msgs in https://github.com/ros-planning/moveit_msgs/pull/14

Modifies moveit motion planning plugin to fill the ROBOT_NAME field in robot_states in the Warehouse.

Also adds python method to find the name of the current robot model.

### DIFF
--- a/perception/semantic_world/CMakeLists.txt
+++ b/perception/semantic_world/CMakeLists.txt
@@ -2,7 +2,7 @@ set(CMAKE_BUILD_TYPE Debug)
 set(MOVEIT_LIB_NAME moveit_semantic_world)
 
 add_library(${MOVEIT_LIB_NAME} src/semantic_world.cpp)
-add_dependencies(${MOVEIT_LIB_NAME} moveit_msgs_generate_messages_cpp )
+add_dependencies(${MOVEIT_LIB_NAME} ${catkin_EXPORTED_TARGETS})
 target_link_libraries(${MOVEIT_LIB_NAME} ${catkin_LIBRARIES} ${Boost_LIBRARIES} ${OpenCV_LIBRARIES})
 
 install(DIRECTORY include/ DESTINATION include)
--- a/perception/semantic_world/CMakeLists.txt
+++ b/perception/semantic_world/CMakeLists.txt
@@ -2,6 +2,7 @@ set(CMAKE_BUILD_TYPE Debug)
 set(MOVEIT_LIB_NAME moveit_semantic_world)
 
 add_library(${MOVEIT_LIB_NAME} src/semantic_world.cpp)
+add_dependencies(${MOVEIT_LIB_NAME} moveit_msgs_generate_messages_cpp )
 target_link_libraries(${MOVEIT_LIB_NAME} ${catkin_LIBRARIES} ${Boost_LIBRARIES} ${OpenCV_LIBRARIES})
 
 install(DIRECTORY include/ DESTINATION include)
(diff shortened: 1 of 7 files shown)


ugocupcic:
Hi, any news regarding this pull request? This is a feature we're really forward to have :)

ugocupcic:
Still no news on this? Could you update us on what might be wrong with the pull request so that we can fix it?

Cheers,

sachinchitta:
Its a lot of new service but overall, LGTM. Would you be ok merging this into jade or do you need it in indigo? Since these are new additions, I think adding to indigo is ok but I will let other people chime in before we do that. @isucan @mikeferguson ?

ugocupcic:
@sachinchitta we're not using jade yet unfortunately :S Merging it indigo would be really great.

planning_interface/robot_interface/src/wrap_python_robot_interface.cpp
```diff
@@ -60,6 +60,11 @@ class RobotInterfacePython : protected py_bindings_tools::ROScppInitializer
     current_state_monitor_ = planning_interface::getSharedStateMonitor(robot_model_, planning_interface::getSharedTF());
   }
 
+  std::string getRobotName() const
```
isucan: I think you should return const char\* here

warehouse/warehouse/src/warehouse_services.cpp
```diff
@@ -0,0 +1,170 @@
...
+#include <moveit_msgs/RenameRobotStateInWarehouse.h>
+
+static const std::string ROBOT_DESCRIPTION="robot_description";
+
+bool storeState(moveit_msgs::SaveRobotStateToWarehouse::Request&  request,
+                 moveit_msgs::SaveRobotStateToWarehouse::Response& response,
+                 moveit_warehouse::RobotStateStorage* rs)
+{
+  const moveit_msgs::RobotState& state = request.state;
+  if ("" == request.name)
```
isucan: Consider using request.name.empty() instead of constructing a new string, running a compare op and then destructing the string.

warehouse/warehouse/src/warehouse_services.cpp
```diff
@@ -0,0 +1,170 @@
...
+  }
+  rs->addRobotState(request.state, request.name, request.robot);
+  return response.success = true;
+}
+
+bool listStates(moveit_msgs::ListRobotStatesInWarehouse::Request&  request,
+                  moveit_msgs::ListRobotStatesInWarehouse::Response& response,
+                  moveit_warehouse::RobotStateStorage* rs)
+{
+  if ("" == request.regex)
```
isucan: ditto

perception/semantic_world/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ set(CMAKE_BUILD_TYPE Debug)
 set(MOVEIT_LIB_NAME moveit_semantic_world)
 
 add_library(${MOVEIT_LIB_NAME} src/semantic_world.cpp)
+add_dependencies(${MOVEIT_LIB_NAME} moveit_msgs_generate_messages_cpp )
```
isucan: I think there was a better way to include this type of deps, but it escapes me right now. @dirk-thomas ?

isucan:
This looks good to me. Please take a look at the small changes that were requested, and I think we can merge.

perception/semantic_world/CMakeLists.txt
dirk-thomas: `${catkin_EXPORTED_TARGETS}` see http://docs.ros.org/api/catkin/html/howto/format1/cpp_msg_dependencies.html

dg-shadow:
Thanks for your comments - I've made the changes you suggested :)

sachinchitta:
https://github.com/ros-planning/moveit_msgs/pull/14 was missed. Merging that now.

PR #595, opened 2015-07-29
Added a max range to raytracing in PointCloudOctomapUpdater

TheBrewCrew:
This will limit how far out the updater will raytrace clipped cells, reducing the amount of time spent raytracing.  This range can be set in the YAML for the updater with the name raytrace_range.  The max range for the filter has been renamed from max_range to obstacle_range and max_range has been deprecated.  Setting raytrace_range to a value slightly larger than obstacle_range is recommended.  I also added functions to parse the YAML with default values so that raytrace_range could default to obstacle_range but decided to make raytrace_range behavior opt in for now so the new functions are not currently used.

### DIFF
--- a/perception/occupancy_map_monitor/include/moveit/occupancy_map_monitor/occupancy_map_updater.h
+++ b/perception/occupancy_map_monitor/include/moveit/occupancy_map_monitor/occupancy_map_updater.h
@@ -107,6 +107,8 @@ protected:
 
   static void readXmlParam(XmlRpc::XmlRpcValue &params, const std::string &param_name, double *value);
   static void readXmlParam(XmlRpc::XmlRpcValue &params, const std::string &param_name, unsigned int *value);
+  static void readXmlParam(XmlRpc::XmlRpcValue &params, const std::string &param_name, double *value, double default_value);
+  static void readXmlParam(XmlRpc::XmlRpcValue &params, const std::string &param_name, unsigned int *value, unsigned int default_value);
 
 };
 
--- a/perception/pointcloud_octomap_updater/include/moveit/pointcloud_octomap_updater/pointcloud_octomap_updater.h
+++ b/perception/pointcloud_octomap_updater/include/moveit/pointcloud_octomap_updater/pointcloud_octomap_updater.h
@@ -81,7 +81,8 @@ private:
   std::string point_cloud_topic_;
   double scale_;
   double padding_;
-  double max_range_;
+  double obstacle_range_;
+  double raytrace_range_;
   unsigned int point_subsample_;
   std::string filtered_cloud_topic_;
   ros::Publisher filtered_cloud_publisher_;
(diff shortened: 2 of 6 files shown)


robot_interaction/include/moveit/robot_interaction/kinematic_options.h
```diff
@@ -62,11 +62,9 @@ struct KinematicOptions
     STATE_VALIDITY_CALLBACK     = 0x00000004, // state_validity_callback_
     LOCK_REDUNDANT_JOINTS       = 0x00000008, // options_.lock_redundant_joints
     RETURN_APPROXIMATE_SOLUTION = 0x00000010, // options_.return_approximate_solution
-    DISCRETIZATION_METHOD       = 0x00000020, // options_.discretization_method
```
mikeferguson, line 65: you should not be modifying these robot_interaction files

perception/occupancy_map_monitor/include/moveit/occupancy_map_monitor/occupancy_map_updater.h
```diff
@@ -107,6 +107,8 @@ class OccupancyMapUpdater
 
   static void readXmlParam(XmlRpc::XmlRpcValue &params, const std::string &param_name, double *value);
   static void readXmlParam(XmlRpc::XmlRpcValue &params, const std::string &param_name, unsigned int *value);
+  static void readXmlParam(XmlRpc::XmlRpcValue &params, const std::string &param_name, double *value, double default_value);
```
mikeferguson, line 110: these actually aren't needed -- if the param does not exist, the value will not be updated (and thus will remain at whatever previous value it had)

mikeferguson, line 110: also don't seem to be used

perception/pointcloud_octomap_updater/include/moveit/pointcloud_octomap_updater/pointcloud_octomap_updater.h
```diff
@@ -81,7 +81,8 @@ class PointCloudOctomapUpdater : public OccupancyMapUpdater
   std::string point_cloud_topic_;
   double scale_;
   double padding_;
-  double max_range_;
+  double obstacle_range_;
```
mikeferguson, line 84: we should add comments here -- noting that these parameters have the same meaning as the obstacle/voxel layers of the navigation stack -- and that obstacle range is the max range at which an obstacle will be inserted, whereas ray trace range is max distance to ray trace free space.

TheBrewCrew:
Opened against wrong branch, should be against indigo-devel instead of jade-devel.  Will fix comments and reopen against correct branch.

mikeferguson:
We'll actually need a PR for both indigo and jade -- only the Jade branch should have the deprecation warning

PR #597, opened 2015-07-30
Add a max range to raytracing in PointCloudOctomapUpdater

TheBrewCrew:
This is the new version of #595 with the changes suggested there and against the correct branch.  A pull request for the same functionality in Jade with a deprecation warning will follow.


TheBrewCrew:
Squashed the commits down to 1 commit

PR #602, opened 2015-08-13
Enable Octomap diff

TheBrewCrew:
This enables the changes in ros-planning/moveit_core#255 by adding a ROS param (/move_group/octomap_diff) to enable change detection on the Octomap.  This pull request should build without ros-planning/moveit_core#255 but will not do anything until it is merged.

PR #604, opened 2015-08-14
Added param to load Octomap from file at startup

TheBrewCrew:
This adds a param (/move_group/octomap_file) that can be set to load a binary Octomap file (.bt) into the Octomap when the MapMonitor starts.  The map can be created using the save_map script in move_group/scripts.

### DIFF
--- a/perception/occupancy_map_monitor/src/occupancy_map_monitor.cpp
+++ b/perception/occupancy_map_monitor/src/occupancy_map_monitor.cpp
@@ -87,6 +87,21 @@ void OccupancyMapMonitor::initialize()
   tree_.reset(new OccMapTree(map_resolution_));
   tree_const_ = tree_;
 
+  std::string map_file = "";
+  if (nh_.getParam("octomap_file", map_file) && !map_file.empty())
+  {
+    ROS_INFO("Reading map from %s", map_file.c_str());
+    try
+    {
+      if (!tree_->readBinary(map_file))
+        ROS_ERROR("Failed to load map from file");
+    }
+    catch (...)
+    {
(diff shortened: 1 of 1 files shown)


perception/occupancy_map_monitor/src/occupancy_map_monitor.cpp
```diff
@@ -303,6 +318,7 @@ bool OccupancyMapMonitor::loadMapCallback(moveit_msgs::LoadMap::Request& request
   tree_->lockWrite();
   try
   {
+    tree_->enableChangeDetection(false);
```
dornhege, line 321: Is there any good reason to explicitly disable change detection here? This seems unrelated to the PR.

TheBrewCrew, line 321: Setting change detection to false is necessary to make this pull request properly interact with pull request #602.

dornhege, line 321: I might be missing something, but why isn't this part of #602 then? It seems unrelated to this feature as it goes into the loadMapCallback and that is already present in the original code.

In general - is this even required? The only problem that I see would be that a tree load with enabled change detection does not reflect the full tree change correctly.

dornhege, line 321: After reviewing ros-planning/moveit_ros#602 (now ros-planning/moveit_ros#636) I believe that this line should not be here (in this PR), but here should actually go to the load via param in line 102.

dornhege:
This merge should be held until after #636 is merged to prevent conflicts and also disable change detection initially if an octree is loaded with this mechanism.

PR #609, opened 2015-09-13
rviz plugin improvements

rhaschke:
This PR allows to save/restore the workspace config in the rviz plugin. Having a non-centered workspace, we ended up, entering those six numbers over and over.
Also, the tab-order in the gui is fixed.


sachinchitta:
Nice change! Again, can you push this into jade too?

PR #611, opened 2015-09-16
fix segfault when disabling and re-enabling TrajectoryVisualization

rhaschke:
animating_path_ was still true causing update() to access displaying_trajectory_message_, which was reset onDisable().


sachinchitta:
This works - mind doing the same against jade-devel?

PR #619, opened 2015-10-09
Fixed compile error caused by BOOST_STATIC_ASSERT in kinematic_options.cpp

daichi-yoshikawa:
Added kinematics::DiscretizationMethods::DiscretizationMethod to QO_FIELDS in kinematic_options.cpp.

At pull request #581, type of discretization_method was set to int. Changed it to proper type.


jonbinney:
+1 for this. Right now I can't compile moveit_ros at all on ubuntu 14.04 because of this problem. @daichi-yoshikawa can you rebase this change to make it mergeable?

jonbinney:
Actually I take that back. My build error was due to trying to build moveit_ros against an older version of moveit_core from debs. Once I got the newest moveit_core and moveit_ros from source, they compiled.

But still, +1 for this change because `::kinematics::DiscretizationMethods::DiscretizationMethod` seems like a more "correct" type here than int.

PR #625, opened 2015-11-18
Kei okada's - get default planner from param

sachinchitta:


sachinchitta:
@rhaschke - this param is only used to set the default planner for when the Rviz plugin gets brought up - currently it comes up as "unspecified". Assigning default planners per group would be useful but it should happen separately from this PR. Maybe we should rename this param to "rviz_default_planner_config" to minimize confusion.

rhaschke:
My adaption loads the `default_planner_config` from the `move_group's` namespace, considering also the selected group when there is a configuration for it. Hence in `ompl_planning.yaml`, one can now specify:

``` yaml
arm:
  default_planner_config: RRTConnectkConfigDefault
  planner_configs:
    - RRTkConfigDefault
    - RRTConnectkConfigDefault
```

Obviously, one can overwrite that setting manually in launch files - after loading the yaml defaults.

The same default could be used by the `move_group` node too. 
However, there is a heuristic to choose a config if I remember correctly.
@k-okada Do you like that modification?

sachinchitta:
Nice change! @rhaschke

PR #628, opened 2015-11-22
rviz gui for planner parameters

rhaschke:
This replaces #627, now based on jade-devel.


sachinchitta:
Overall, looks great. Is there a way to wrap the planner parameter names (keys) so they are more fully visible? Also, the vertical line in the display does not seem to serve a purpose - could it be setup to actually separate the keys and values?

Again, great contribution.

rhaschke:
Huh. Wrapping parameter names would be fancy. However, I wouldn't go for it.
Rather, I fixed an issue in rviz related to the vertical line. [See rviz PR #943](https://github.com/ros-visualization/rviz/pull/943). With that fix, the line will work as expected.

PR #652, opened 2016-03-11
Update frame transforms in PlanningScene callback as well

marcoesposito1988:
Sending a PlanningScene message which adds object to the environment does not trigger a call to updateFrameTransforms. The result can be an annoying error of the type as in the following topic:

http://answers.ros.org/question/166606/how-do-i-use-collision_objectheaderframe_id-can-i-use-it-for-frame-transformations/

Is there a reason why PlanningScene does not trigger it, while all other topic callbacks do? If not, this is a straight-forward bugfix.

### DIFF
--- a/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
+++ b/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
@@ -480,6 +480,7 @@ void planning_scene_monitor::PlanningSceneMonitor::newPlanningSceneMessage(const
 {
   if (scene_)
   {
+    updateFrameTransforms();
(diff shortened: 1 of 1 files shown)


v4hn:
Good question. I don't know of a reason not to update here.
Actually, this issue is the reason why I added an explicit updateFrameTransforms here: https://github.com/ros-planning/moveit_ros/commit/806ca363c681c6bec4f644ec2b222aa4233d708d#diff-10290b25cee00537f3a1c5004bf4f711R57

+1 on adding the update to newPlanningScene instead.
I'll open a new request as this one cannot be applied anymore.

Issue #654, opened 2016-03-23
fails to compile with the latest version of rviz


wjwwood:
For example:

http://build.ros.org/view/Ibin_uT64/job/Ibin_uT64__moveit_ros_benchmarks_gui__ubuntu_trusty_amd64__binary/10/console

I believe the root cause of this is that the "default_plugin" is no longer included in rviz's `_LIBRARIES` variable, but `moveit_ros_benchmark_gui` uses headers from the default plugin and so it needs to link against them. There is now a cmake variable called `rviz_DEFAULT_PLUGIN_LIBRARIES` which you can add to your `target_link_libraries` which will contain the default plugin. I recommend that this variable be used to fix the current compilation problems.

See:
- https://github.com/ros-visualization/rviz/pull/948
- https://github.com/ros-visualization/rviz/issues/964
- https://github.com/ros-visualization/rviz/pull/979

dornhege:
Partially solved with #657 and #658. I'm not sure if this can be solved for catkin_make without any changes in rviz.

wjwwood:
Can anyone try with https://github.com/ros-visualization/rviz/pull/983?

dornhege:
This doesn't fix it for me. The problem is that the rviz extras wants the generated file and thus fails before linking during cmake already.

```
CMake Error at .../devel/share/rviz/cmake/rviz-extras.cmake:2 (include):
  include could not find load file:
.../devel/share/rviz/cmake/default_plugin_location.cmake
Call Stack (most recent call first):
  .../devel/share/rviz/cmake/rvizConfig.cmake:190 (include)
  /opt/ros/indigo/share/catkin/cmake/catkinConfig.cmake:75 (find_package)
moveit_ros/benchmarks_gui/CMakeLists.txt:19 (find_package)
```

wjwwood:
I see, let me try to address that.

wjwwood:
I added https://github.com/ros-visualization/rviz/pull/983/commits/91ba11a4bc18c55a857a10e61e975e8cebc8c052 to try and address that. Can you give that a try?

dornhege:
It did compile with catkin_make. However only after multiple runs with `--force-cmake`. On the first run I'm getting linker errors, because it's not linking the library.

wjwwood:
@dornhege is that from a clean build that it fails on the first pass?

dornhege:
Yes, after I removed build/ and devel/.

wjwwood:
[...]

PR #657, opened 2016-03-29
explicitly link rviz' default_plugin library

v4hn:
The library is not exported anymore and now is provided separately from
rviz_LIBRARIES. See https://github.com/ros-visualization/rviz/pull/979 for
details.

PR #658, opened 2016-03-29
explicitly link rviz' default_plugin library - indigo request

v4hn:
This is the indigo version of #657. Feel free to cherry-pick
the commit to indigo instead of merging this request.

The library is not exported anymore and now is provided separately from
rviz_LIBRARIES. See https://github.com/ros-visualization/rviz/pull/979 for
details.

### DIFF
--- a/benchmarks_gui/CMakeLists.txt
+++ b/benchmarks_gui/CMakeLists.txt
@@ -70,6 +70,7 @@ add_executable(moveit_benchmark_gui ${SOURCES} ${MOC_FILES} ${UIC_FILES})
 
 target_link_libraries(moveit_benchmark_gui
   ${catkin_LIBRARIES}
+  ${rviz_DEFAULT_PLUGIN_LIBRARIES}
   ${OGRE_LIBRARIES}
   ${QT_LIBRARIES}
   ${Boost_LIBRARIES}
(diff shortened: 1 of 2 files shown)


k-okada:
This fixes #654, merge please!

dornhege:
This does not seem to build with catkin_make (but with catkin build), because the generated target cmake file is build at compile time, but required at cmake configuration time.

I will merge this now as it is an improvement and should remove the regression in the build farm.

v4hn:
@dornhege: you're right. This only resolves the problem with catkin_make_isolated
and for cases where rviz is already installed before compiling moveit_ros.

The `CFG_EXTRAS` mechanism @wjwwood pulled out of his magic hat seems to be broken w.r.t. that.
Maybe one could also export the real library target as `rviz_DEFAULT_PLUGIN_LIBRARIES` in rviz' CMakeLists.txt?
I'm not sure there's a clean solution for the general issue though.

Anyways, this solves _most_ of the use-cases. rviz is usually only
in the core system that _must_ be built with catkin_build_isolated afair.

Thanks for merging.

wjwwood:
I can put the default plugins in the exported targets maybe. I'd have to think about that some more.

The problem with getting the library name at configure time is that CMake doesn't actually know it, or in the future it won't, until generation is done, which is why they introduced generator expressions.

wjwwood:
Actually I think I can just put the target in the `rviz_DEFAULT_PLUGIN_LIBRARIES` cmake variable within the rviz `CMakeLists.txt`. That way with `catkin_make` the `rviz_DEFAULT_PLUGIN_LIBRARIES` will be defined and point to the right target.

wjwwood:
I opened https://github.com/ros-visualization/rviz/pull/983 as a potential solution.

PR #660, opened 2016-03-30
add rviz_DEFAULT_PLUGIN_LIBRARIES

### LINKED ISSUE #654: fails to compile with the latest version of rviz

k-okada:
this is indio version of  #657, fixes #654
see https://github.com/ros-visualization/rviz/issues/964

### DIFF
--- a/benchmarks_gui/CMakeLists.txt
+++ b/benchmarks_gui/CMakeLists.txt
@@ -70,6 +70,7 @@ add_executable(moveit_benchmark_gui ${SOURCES} ${MOC_FILES} ${UIC_FILES})
 
 target_link_libraries(moveit_benchmark_gui
   ${catkin_LIBRARIES}
+  ${rviz_DEFAULT_PLUGIN_LIBRARIES}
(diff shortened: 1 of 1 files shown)


k-okada:
sorry this is duplicates of #658

PR #664, opened 2016-04-23
[perception] Find GL library on OS X

wkentaro:


rhaschke:
Setting hard-coded paths isn't a good idea. Why does

``` cmake
find_package(OpenGL REQUIRED)
find_package(GLEW REQUIRED)
find_package(GLUT REQUIRED)
```

succeed, but doesn't set the appropriate variables? 
Please check the cmake files of OpenGL / GLUT and fix them instead.

wkentaro:
@rhaschke 
Thank you for the quick reply. I have worked around and reached the solution as the commit.
To build this package on OS X (10.11), it is necessary to find X11.

### Reasons
- `-lglut` not found error

```
% locate libglut
/opt/X11/lib/libglut.3.dylib
/opt/X11/lib/libglut.dylib
```
- `GL/glut.h` not found error

```
% locate GL/glut.h
/Developer/NVIDIA/CUDA-7.5/extras/CUPTI/include/GL/glut.h
/Developer/NVIDIA/CUDA-7.5/samples/common/inc/GL/glut.h
/opt/X11/include/GL/glut.h
```

perception/CMakeLists.txt
```diff
@@ -74,10 +77,12 @@ include_directories(SYSTEM
                     ${Boost_INCLUDE_DIRS}
                     ${GLEW_INCLUDE_DIR}
                     ${GLUT_INCLUDE_DIR}
+                    ${X11_INCLUDE_DIR}
                     )
 
 link_directories(${Boost_LIBRARY_DIRS})
 link_directories(${catkin_LIBRARY_DIRS})
+link_directories(${X11_LIBRARY_DIR})
```
rhaschke, line 85: Better. Could you please try to replace ${X11_INCLUDE_DIR} by ${OPENGL_INCLUDE_DIR}?

wkentaro, line 85: The value of `OPENGL_INCLUDE_DIR` is below, so it does not work. (GL/glut.h
is not found).

/System/Library/Frameworks/OpenGL.framework

rhaschke:
I see. However, as this is obviously a bug upstream (in package OpenGL or glut), can I ask you to try another thing:
As rviz uses the same `find_package(OpenGL)` without any `APPLE`-specific settings, could you please try to build rviz and report on the results there?

wkentaro:
Sorry for late reply.
It seems that `GL/glut.h` is not used in rviz.

v4hn:
Maybe I'm missing something, but using `link_directories` is discouraged in ROS (`catkin_lint` complains).
Now this request added a new `link_directories` just recently.

Is there no official way to add folders of dynamic libraries to the search path in OS X?
Do we _have_ to provide these `link_directories` commands to support apple?
Just last week I removed some of them from [moveit_core](https://github.com/ros-planning/moveit_core/commit/9d909ad5c294af328f91d63b5ecaaf1e34eea85c).

PR #669, opened 2016-05-04
Fix broken frame transforms in moveit's rviz displays

v4hn:
This makes using the displays with rviz a much nicer experience. :)

Now all moveit displays behave correctly when changing the fixed-frame
in rviz to an end-effector.

The more common use-case where problems occurred is when
moveit's planning frame is base_link of a mobile base,
but the fixed-frame is map. Because this transform usually
changes as the robot localizes itself, assuming a static transform
is broke.

Please cherry-pick and release this fix into indigo as well!
(If you prefer I can also create an additional pull request for that..)

### DIFF
--- a/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp
+++ b/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp
@@ -200,7 +200,7 @@ void MotionPlanningDisplay::onInitialize()
   PlanningSceneDisplay::onInitialize();
 
   // Planned Path Display
-  trajectory_visual_->onInitialize(scene_node_, context_, update_nh_);
+  trajectory_visual_->onInitialize(planning_scene_node_, context_, update_nh_);
 
   query_robot_start_.reset(new RobotStateVisualization(planning_scene_node_, context_, "Planning Request Start", NULL));
   query_robot_start_->setCollisionVisible(false);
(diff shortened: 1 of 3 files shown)


rhaschke:
Thanks for this fix.

rhaschke:
While this series of fixes handles the correct display of the PlanningScene in rviz, I think there is a more fundamental problem: The frame_id of collision_objects is only used once at instantiation time. Later (tf) updates to the frames are ignored - in the PlanningScene everything is expressed w.r.t. the root model frame.
Hence, if you have placed collision objects relative to `map` (because you know they are fixed), they will move together with the robot, when the transform from map to robot is updated.

@isucan Is there a TF-Updater plugin, which takes care of transform updates in the planning scene when tf frames do change?

v4hn:
On Thu, May 19, 2016 at 07:51:14AM -0700, Robert Haschke wrote:

> I think there is a more fundamental problem: The frame_id of collision_objects
> is only used once at instantiation time. Later (tf) updates to the frames are ignored
> - in the PlanningScene everything is expressed w.r.t. the root model frame.

Yes. I am rather sure this is deliberate and more of a modelling decision than a problem.
To simplify computation, moveit works in its ModelFrame at all times.
Therefore it expects things not to change with respect to it unless one MOVEs them.
I would say, if you want your objects to stay constant w.r.t. a map frame, make it the ModelFrame.
In theory one could maintain a list of frame-ids for the objects and update them,
or postpone the transformation to the collision world.
I'm not sure it's worth the effort though.
Also, I think in this case there should be a mechanism such as `visualization_msgs/Marker`'s `frame_locked` added.
Otherwise, all objects you add in the mobile robot's camera frame would move around with the robot and that's probably not a good idea either.

Anyways, I'd say this is an independent feature request for moveit
and has nothing really to do with this request that fixed the rviz plugins.

Issue #676, opened 2016-05-23
The catkin_package LIBRARIES list is missing moveit_collision_plugin_loader


Levi-Armstrong:
This is true for both indigo-devel and jade-devel branches.

[Library](https://github.com/ros-planning/moveit_ros/blob/jade-devel/planning/collision_plugin_loader/CMakeLists.txt#L1)
[Missing here](https://github.com/ros-planning/moveit_ros/blob/jade-devel/planning/CMakeLists.txt#L47-L55)

rhaschke:
Actually, the library is implicitly linked into [moveit_planning_scene_monitor](https://github.com/ros-planning/moveit_ros/blob/jade-devel/planning/planning_scene_monitor/CMakeLists.txt#L9), which in turn [is listed in catkin_package](https://github.com/ros-planning/moveit_ros/blob/jade-devel/planning/CMakeLists.txt#L55).
What is the actual error you get?

Levi-Armstrong:
I am getting the error: undefined reference to `collision_detection::CollisionPluginLoader::CollisionPluginLoader()'. I downloaded the source and made the change and it compiled without any errors. I did not try running it with the source without the change, but I tried it this morning and it also compiled without errors.

So my code compiling using the source but not with the binary.

rhaschke:
Could you please provide more context information:
What are you compiling? Whole MoveIt workspace? Parts of it? Your own project?
Which build environment do you use? OS? Compiler? Linker?
As you can see on [Travis CI](https://travis-ci.org/ros-planning/moveit_ros), `moveit_ros` compiles without issues from source.

Levi-Armstrong:
> What are you compiling?

I am compiling a custom package which is using CollisionPluginLoader. Because of the compiler errors I then cloned the moveit_ros repository and compiled the whole workspace which includes my custom package and moveit_ros and the compiler errors went away.

> Which build environment do you use? OS? Compiler? Linker?

I am using catkin_make on Ubuntu 14.04.
gcc version 4.8.4 (Ubuntu 4.8.4-2ubuntu1~14.04.3) 

> As you can see on Travis CI, moveit_ros compiles without issues from source.

The issue is not with the source, but possible with the currently released binary. It is possible that it is something with my package but if it was I would have expected it not to compile using moveit_ros source also.

Levi-Armstrong:
@rhaschke ,

[...]

PR #677, opened 2016-05-24
Add library moveit_collision_plugin_loader as an exported catkin library

Levi-Armstrong:
For other catkin package to link against library moveit_collision_plugin_loader it must be added as a catkin library to be exported.


Levi-Armstrong:
This address issue #676 for indigo-devel branch.

PR #683, opened 2016-06-07
cleanup cmake tests, fix empty output

mikeferguson:
See discussion in #679


rhaschke:
+1. Probably, `rostest` should be mentioned in [package.xml](http://wiki.ros.org/catkin/package.xml#Dependencies)'s `<test_depend>` too.

mikeferguson:
@rhaschke -- good catch -- It was added as a build depend previously, that is incorrect -- I'm going to fixup and merge

PR #685, opened 2016-06-13
clear_octomap_service: fix runtime name

v4hn:
Looks like the author copy&pasted from a different capability
and forgot to change the name.


rhaschke:
+1

PR #688, opened 2016-06-14
planning_interface: Set is_diff true for empty start state

Jntzko:
Executing a motion without setting the start state of the robot like
this:

```
group_arm.setNamedTarget("start_grab_pose");
success = group_arm.move();
```

throws the error: Execution of motions should always start at the robot's
current state. Ignoring the state supplied as start state in the motion
planning request.

The problem is, when considered_start_state_ is null, every data field of the start_state
in the submitted MotionPlanRequest is 0 or false. But we need is_diff to be
true, because otherwise move_group will not consider its current state as
actual start state without complaining.

### DIFF
--- a/planning_interface/move_group_interface/src/move_group.cpp
+++ b/planning_interface/move_group_interface/src/move_group.cpp
@@ -856,6 +856,8 @@ public:
 
     if (considered_start_state_)
       robot_state::robotStateToRobotStateMsg(*considered_start_state_, goal.request.start_state);
+    else
+      goal.request.start_state.is_diff = true;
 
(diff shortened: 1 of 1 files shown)


v4hn:
As I supervised Jntzko writing the patch, I'll not merge it myself for now.

@Jntzko: The same `if(considered_start_state_)`` construct is
present in`computeCartesianPath``` in the same file,
could you please update it there too?

Jntzko:
@v4hn I updated this too.

davetcoleman:
+1 this makes sense to me

PR #693, opened 2016-06-23
Shadow indigo latest

toliver:
- This PR adds methods to retrieve named  states via the move_group class.
- It also adds the corresponding python bindings for these methods.
- I also allows for the warehouse DB host to be specified from a parameter.


v4hn:
Heya,

Actually I was thinking of adding this request
myself after I found out Shadow has their own
patchset which they provide for their users.
So a big +1 to have this request here.

However, the branch, as it is, is quite a mess...

(1) Could you please use `git rebase` to get rid of
all the merge commits? This is pretty straight forward
and I already did this once locally last month...

(2) Could you please squash the fixup commits
together with the commits introducing the code?
At least 0f1a9f7 ,  5f9bbac and  c4a1f51
should definitely be only one commit.

The same goes for  056bf8b and 997d8ba .

v4hn:
sorry, I mixed up commit ids, just updated the ids in the previous comment.

toliver:
Hi @v4hn. Yes, sorry for the mess. I wasn't sure if this PR would be of interest. I'm glad it is. I will tidy up the commits.

planning_interface/move_group_interface/src/move_group.cpp
```diff
@@ -200,6 +201,7 @@ class MoveGroup::MoveGroupImpl
       constraints_init_thread_->join();
   }
 
+
```
davetcoleman, line 204: outdated minor changes like this should probably be cleaned up

planning_interface/move_group_interface/src/move_group.cpp
```diff
@@ -1094,6 +1096,24 @@ const std::string& moveit::planning_interface::MoveGroup::getName() const
   return impl_->getOptions().group_name_;
 }
 
+const std::vector<std::string> moveit::planning_interface::MoveGroup::getNamedTargets()
+{
+  std::vector<std::string> output;
+  const robot_model::RobotModelConstPtr& robot = getRobotModel();
+  std::string group = getName();
```
davetcoleman, line 1103: const std::string& ?

planning_interface/move_group_interface/src/move_group.cpp
```diff
@@ -1094,6 +1096,24 @@ const std::string& moveit::planning_interface::MoveGroup::getName() const
...
+  const robot_model::JointModelGroup* joint_group = robot->getJointModelGroup(group);
+
+  if (joint_group)
+  {
+    const std::vector<std::string>& known_states = joint_group->getDefaultStateNames();
+    if (!known_states.empty())
+    {
+      output.insert(output.end(), known_states.begin(), known_states.end());
```
davetcoleman, line 1111: im not sure, but is it necessary to copy this vector into a temporary vector, then return that vector?

visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp
```diff
@@ -1248,10 +1248,17 @@ void MotionPlanningDisplay::load(const rviz::Config& config)
...
+    if (nh.getParam("warehouse_host", host_param))
+    {
+      host = QString::fromStdString(host_param);
+      frame_->ui_->database_host->setText(host);
+    }
+    else if (config.mapGetString("MoveIt_Warehouse_Host", &host))
       frame_->ui_->database_host->setText(host);
     int port;
-    if (config.mapGetInt("MoveIt_Warehouse_Port", &port))
+    if (nh.getParam("warehouse_port", port) || config.mapGetInt("MoveIt_Warehouse_Port", &port))
```
davetcoleman, line 1261: this appears to be a breaking change in functionality in that the way settings are loaded are different. i think this PR would need to target ROS Kinetic branch

v4hn, line 1261: The requests consists of multiple orthogonal patches.
I do believe most of them can be merged into indigo without any problem.
Let's see what it looks like once it's cleaned. :-)

Concerning this particular patch, I do not believe it is actually a reasonable
behavior to ignore the item in the rviz config if the rosparam is available,
so this part should probably be changed anyway. It's probably easy to make it
backward-compatible. E.g. one could fill in the values only when they are not
specified/empty in the config.

toliver:
Closing. Rebased version in #694

PR #709, opened 2016-07-10
Stop execution on motion planning rviz plugin

wkentaro:
moveit_motion_planning_rviz_plugin_stop_button

### DIFF
--- a/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
+++ b/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
@@ -65,6 +65,12 @@ void MotionPlanningFrame::planAndExecuteButtonClicked()
   planning_display_->addBackgroundJob(boost::bind(&MotionPlanningFrame::computePlanAndExecuteButtonClicked, this), "plan and execute");
 }
 
+void MotionPlanningFrame::stopButtonClicked()
+{
+  ui_->stop_button->setEnabled(false);
+  planning_display_->addBackgroundJob(boost::bind(&MotionPlanningFrame::computeStopButtonClicked, this), "stop");
+}
+
 void MotionPlanningFrame::allowReplanningToggled(bool checked)
 {
   if (move_group_)
@@ -129,7 +135,22 @@ void MotionPlanningFrame::computePlanButtonClicked()
 void MotionPlanningFrame::computeExecuteButtonClicked()
 {
   if (move_group_ && current_plan_)
-    move_group_->execute(*current_plan_);
+    ui_->stop_button->setEnabled(true);
+    move_group_->asyncExecute(*current_plan_);
+}
+
+void MotionPlanningFrame::computeStopButtonClicked()
+{
+  if (move_group_) {
+    current_plan_.reset();
+    move_group_->stop();
+    move_group_->clearPoseTargets();
+    move_group_->setJointValueTarget(move_group_->getCurrentJointValues());
+    move_group_->move();
+
+    // Stop
+    ui_->result_label->setText("Stopped");
+  }
 }
 
 void MotionPlanningFrame::computePlanAndExecuteButtonClicked()
@@ -137,7 +158,8 @@ void MotionPlanningFrame::computePlanAndExecuteButtonClicked()
   if (!move_group_)
     return;
   configureForPlanning();
-  move_group_->move();
+  ui_->stop_button->setEnabled(true);
+  move_group_->asyncMove();
   ui_->plan_and_execute_button->setEnabled(true);
 }
 
--- a/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
+++ b/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
@@ -135,20 +135,17 @@ void MotionPlanningFrame::computePlanButtonClicked()
 void MotionPlanningFrame::computeExecuteButtonClicked()
 {
   if (move_group_ && current_plan_)
+  {
     ui_->stop_button->setEnabled(true);
     move_group_->asyncExecute(*current_plan_);
+  }
 }
 
 void MotionPlanningFrame::computeStopButtonClicked()
 {
-  if (move_group_) {
-    current_plan_.reset();
+  if (move_group_)
+  {
     move_group_->stop();
-    move_group_->clearPoseTargets();
-    move_group_->setJointValueTarget(move_group_->getCurrentJointValues());
-    move_group_->move();
-
-    // Stop
     ui_->result_label->setText("Stopped");
   }
 }
--- a/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
+++ b/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
@@ -40,6 +40,7 @@
 #include <moveit/kinematic_constraints/utils.h>
 #include <moveit/robot_state/conversions.h>
 
+#include <actionlib/client/simple_action_client.h>
 #include <std_srvs/Empty.h>
 
 #include "ui_motion_planning_rviz_plugin_frame.h"
@@ -55,6 +56,7 @@ void MotionPlanningFrame::planButtonClicked()
 void MotionPlanningFrame::executeButtonClicked()
 {
   ui_->execute_button->setEnabled(false);
+  ui_->plan_and_execute_button->setEnabled(false);
   planning_display_->addBackgroundJob(boost::bind(&MotionPlanningFrame::computeExecuteButtonClicked, this), "execute");
 }
 
@@ -137,7 +139,7 @@ void MotionPlanningFrame::computeExecuteButtonClicked()
   if (move_group_ && current_plan_)
   {
     ui_->stop_button->setEnabled(true);
-    move_group_->asyncExecute(*current_plan_);
+    move_group_->asyncMove();
   }
 }
 
@@ -404,5 +406,18 @@ void MotionPlanningFrame::remoteUpdateGoalStateCallback(const std_msgs::EmptyCon
   }
 }
 
-
+void MotionPlanningFrame::updatePlanExecutionStatus(float wall_dt, float ros_dt)
+{
+  if (move_group_)
+  {
+    actionlib::SimpleClientGoalState goal_state = move_group_->getMoveActionClientState();
+    if (goal_state != actionlib::SimpleClientGoalState::ACTIVE)
+    {
+      ui_->plan_button->setEnabled(true);
+      ui_->plan_and_execute_button->setEnabled(true);
+      ui_->stop_button->setEnabled(false);
+    }
+  }
 }
+
+}  // namespace moveit_rviz_plugin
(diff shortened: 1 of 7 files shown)


davetcoleman:
Sounds cool - can you provide a screenshot of the GUI change?

wkentaro:
I added a GIF.

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
```diff
@@ -129,15 +135,31 @@ void MotionPlanningFrame::computePlanButtonClicked()
 void MotionPlanningFrame::computeExecuteButtonClicked()
 {
   if (move_group_ && current_plan_)
-    move_group_->execute(*current_plan_);
+    ui_->stop_button->setEnabled(true);
+    move_group_->asyncExecute(*current_plan_);
```
davetcoleman: you need brackets around these two lines

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
```diff
@@ -129,15 +135,31 @@ void MotionPlanningFrame::computePlanButtonClicked()
...
+}
+
+void MotionPlanningFrame::computeStopButtonClicked()
+{
+  if (move_group_) {
```
davetcoleman: this bracket should be on the next line per ros style

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
```diff
@@ -129,15 +135,31 @@ void MotionPlanningFrame::computePlanButtonClicked()
...
+    current_plan_.reset();
+    move_group_->stop();
+    move_group_->clearPoseTargets();
+    move_group_->setJointValueTarget(move_group_->getCurrentJointValues());
+    move_group_->move();
```
davetcoleman: Why is stop() not enough? Does it not cancel the ActionServerRequest which should stop execution?

wkentaro:
Oops, I sent the wip code. Fixed.

davetcoleman:
+1, looks great!

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
```diff
@@ -129,15 +135,28 @@ void MotionPlanningFrame::computePlanButtonClicked()
...
 }
 
 void MotionPlanningFrame::computePlanAndExecuteButtonClicked()
 {
   if (!move_group_)
     return;
   configureForPlanning();
-  move_group_->move();
+  ui_->stop_button->setEnabled(true);
+  move_group_->asyncMove();
```
rhaschke, line 161: Changing from synchronous to asynchronous command here, immediately (re)enables the "plan+execute" button.
This changes behaviour. Before, the button was enabled only when `move()` finished.

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
```diff
@@ -129,15 +135,28 @@ void MotionPlanningFrame::computePlanButtonClicked()
 void MotionPlanningFrame::computeExecuteButtonClicked()
 {
   if (move_group_ && current_plan_)
-    move_group_->execute(*current_plan_);
+  {
```
rhaschke, line 140: Do you really need to run `computeExecuteButtonClicked()` and `computeStopButtonClicked()` in a separate thread (as a background job)? As far as I can see, these methods require only short ROS service call interaction.
Directly executing those commands in `executeButtonClicked()` and `stopButtonClicked()` would simplify the code.
However, the ROS service might be not available.

rhaschke:
Nice feature. Have a look at the inline comments.

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
wkentaro, line 140: Do u mean `computeExecuteButtonClicked()` and `computeStopButtonClicked()` seems to take few time to execute, and it seems not necessary to run it in a separate thread?

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
wkentaro, line 161: Maybe reenabling the button after the execution is necessary?

rhaschke, line 161: The button should be only re-enabled after execution. Because you switch to asynchronous execution now, the button gets re-enabled immediately.

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
rhaschke, line 140: Exactly.

v4hn, line 140: Please keep the background jobs in there.
This is common across RViz and whenever RViz hangs when operating a robot over wifi, I can be sure that someone did _not_ start a background job.
In that case the problem is not the computation time of the RPC, but the network latency involved.

rhaschke, line 140: Valid point.

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
wkentaro, line 161: Fixed.

rhaschke:
Hm. While your recent code will fix the enabling of the stop button, I think the code is rather complicated.
Why do you need to use asynchronous execution at all? The `execute()` and `move()` calls were already executed within a separate thread of rviz. Hence, there is no problem to synchronously wait for their end.
Did you tried that?

wkentaro:
> Did you tried that?

Yeah, I have tried `move()` calls and tried to pushing the stop button, but the joint trajectory action could not be canceled.

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
```diff
@@ -129,15 +137,28 @@ void MotionPlanningFrame::computePlanButtonClicked()
 void MotionPlanningFrame::computeExecuteButtonClicked()
 {
   if (move_group_ && current_plan_)
-    move_group_->execute(*current_plan_);
```
rhaschke, line 132: Replacing `move_group_->execute()` with `move_group_->asyncMove()` is not possible, because behaviour is changed: While the former executes a previously planned, manually validated trajectory, the latter will actually plan a new trajectory and directly execute it - with no chance for inspection/validation.

**I cannot accept this PR in this form.** 

Looking deeper into the reasons, I found the following: There are essentially to methods to plan/execute:
- `move_group::plan()` and `move_group::move()` are calls to an _action_, that can be interrupted. The flag `plan_only` allows to switch off/on execution.
- However, `move_group::execute()` is a call to a service (`MoveGroupExecuteService`), which cannot be interrupted.

In order to allow stopping of a trajectory, the service should be turned into an action as well. As far as I can see, that changed could be hidden transparently behind the existing MoveGroupInterface. However, I'm not sure, whether some (3rd-party) code is directly calling the service. Within the MoveIt code base, I couldn't find any other usage.

@davetcoleman Do you think, this change is reasonable?

Alternatively, we do not allow to stop an executed-only trajectory.

wkentaro, line 132: > In order to allow stopping of a trajectory, the service should be turned into an action as well. 

Precisely speaking, for stopping of a trajectory, the service call with `wait_for_execution:=false` and `stop()` method call is enough. The reason I needed to change the `execute()` method to `move()` is to get the trajectory execution server status to handle GUI button status correctly in https://github.com/ros-planning/moveit_ros/pull/709/commits/5503fa925adbf5a795d4b5c7095117ccb1724f03.

wkentaro, line 132: > Replacing move_group_->execute() with move_group_->asyncMove() is not possible, because behaviour is changed: While the former executes a previously planned, manually validated trajectory, the latter will actually plan a new trajectory and directly execute it - with no chance for inspection/validation.

I see.

>  the service should be turned into an action as well.

I agree with the change.

rhaschke:
If you agree, please close this PR in favor of #713.

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
rhaschke, line 132: > Precisely speaking, for stopping of a trajectory, the service call 
> with |wait_for_execution:=false| and |stop()| method call is enough. 
> The reason I needed to change the |execute()| method to |move()| is to 
> get the trajectory execution server status to handle GUI button status 
> correctly in 5503fa9
> The service call doesn't have a stop() method, does it? I guess you are 
> intermixing with the move action?

rhaschke, line 132: @wkentaro: Could you propose a PR (w.r.t. #713) for turning the execute service into an action?
Dave[ agreed on that approach](https://github.com/ros-planning/moveit_ros/pull/713#issuecomment-232848243), but I don't have time for it right now.
I think, having that done too, we can merge this branch.

wkentaro, line 132: > The service call doesn't have a stop() method, does it? I guess you are
> intermixing with the move action?

I don't use `stop()` method of the move action, but use `stop()` method in `move_group` with trajectory_event_manager.
https://github.com/ros-planning/moveit_ros/blob/kinetic-devel/planning_interface/move_group_interface/src/move_group.cpp#L763
https://github.com/ros-planning/moveit_ros/blob/kinetic-devel/planning_interface/move_group_interface/src/move_group.cpp#L130

I think the execution can be interrupted if `wait_for_execution=false` for execution service, with `stop()` calling in move_group.

rhaschke, line 132: Indeed. That looks promising. Could you prepare a PR for this?

wkentaro, line 132: > Indeed. That looks promising. Could you prepare a PR for this?

This PR without [last commit](https://github.com/ros-planning/moveit_ros/pull/709/commits/5503fa925adbf5a795d4b5c7095117ccb1724f03) is that approach. The trajectory execution with execute_button on rviz plugin was able to be interrupted with `asyncExecute()` and `move_group_->stop()`.
However, it is unable to get the status of execution of service call, so I could not handle the GUI button status, enable/disable, correctly.
That's why I needed to change the `asyncExecute()` to `asyncMove()` in the last commit, in order to handle the enable/disable status of GUI precisely.

rhaschke, line 132: @wkentaro Doesn't work like you described. Of course, using asynchronous execution (`wait_for_execution=false`) will allow the stop call to get through and stop motion.
However, asynchronous execution through a service doesn't allow to get feedback when execution finished. We need to turn the service into an action.

wkentaro, line 132: Yeah. You're right. I meant it does _not_ work.

rhaschke:
Closed in favor of #713.

PR #713, opened 2016-07-15
rviz plugin: stop execution + update of start state to current

### LINKED ISSUE #646: Set the start planning state as "current" after finish the execution of a trajectory

rhaschke:
This PR combines ideas of #709 and #710 (and would replaces those PRs) to allow
- stopping of an executed trajectory (not working for execute-only button, see [comment](https://github.com/ros-planning/moveit_ros/pull/709#discussion_r70888457) in #709)
- updating the start state after execution, if current state is selected in gui (fixes #646)

I consider this work-in-progress. @davetcoleman Should we address [this comment](https://github.com/ros-planning/moveit_ros/pull/709#discussion_r70888457) like suggested?

### DIFF
--- a/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
+++ b/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
@@ -65,6 +65,12 @@ void MotionPlanningFrame::planAndExecuteButtonClicked()
   planning_display_->addBackgroundJob(boost::bind(&MotionPlanningFrame::computePlanAndExecuteButtonClicked, this), "plan and execute");
 }
 
+void MotionPlanningFrame::stopButtonClicked()
+{
+  ui_->stop_button->setEnabled(false); // avoid clicking again
+  planning_display_->addBackgroundJob(boost::bind(&MotionPlanningFrame::computeStopButtonClicked, this), "stop");
+}
+
 void MotionPlanningFrame::allowReplanningToggled(bool checked)
 {
   if (move_group_)
@@ -129,9 +135,11 @@ void MotionPlanningFrame::computePlanButtonClicked()
 void MotionPlanningFrame::computeExecuteButtonClicked()
 {
   if (move_group_ && current_plan_)
-    move_group_->execute(*current_plan_);
-
-  updateStartStateToCurrent();
+  {
+    ui_->stop_button->setEnabled(true); // enable stopping
+    bool success = move_group_->execute(*current_plan_);
+    onFinishedExecution(success);
+  }
 }
 
 void MotionPlanningFrame::computePlanAndExecuteButtonClicked()
@@ -139,18 +147,31 @@ void MotionPlanningFrame::computePlanAndExecuteButtonClicked()
   if (!move_group_)
     return;
   configureForPlanning();
-  move_group_->move();
+  ui_->stop_button->setEnabled(true);
+  bool success = move_group_->move();
+  onFinishedExecution(success);
   ui_->plan_and_execute_button->setEnabled(true);
-
-  updateStartStateToCurrent();
 }
 
-void MotionPlanningFrame::updateStartStateToCurrent()
+void MotionPlanningFrame::computeStopButtonClicked()
 {
+  if (move_group_)
+    move_group_->stop();
+}
+
+void MotionPlanningFrame::onFinishedExecution(bool success)
+{
+  // visualize result of execution
+  if (success) 
+    ui_->result_label->setText("Executed");
+  else 
+    ui_->result_label->setText(!ui_->stop_button->isEnabled() ? "Stopped" : "Failed");
+  // disable stop button
+  ui_->stop_button->setEnabled(false);
+
+  // update query start state to current if neccessary
   if (ui_->start_state_selection->currentText() == "<current>")
-  {
     useStartStateButtonClicked();
-  }
 }
 
 void MotionPlanningFrame::useStartStateButtonClicked()
--- a/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
+++ b/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
@@ -130,6 +130,8 @@ void MotionPlanningFrame::computeExecuteButtonClicked()
 {
   if (move_group_ && current_plan_)
     move_group_->execute(*current_plan_);
+
+  updateStartStateToCurrent();
 }
 
 void MotionPlanningFrame::computePlanAndExecuteButtonClicked()
@@ -139,6 +141,16 @@ void MotionPlanningFrame::computePlanAndExecuteButtonClicked()
   configureForPlanning();
   move_group_->move();
   ui_->plan_and_execute_button->setEnabled(true);
+
+  updateStartStateToCurrent();
+}
+
+void MotionPlanningFrame::updateStartStateToCurrent()
+{
+  if (ui_->start_state_selection->currentText() == "<current>")
+  {
+    useStartStateButtonClicked();
+  }
 }
 
 void MotionPlanningFrame::useStartStateButtonClicked()
--- a/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
+++ b/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
@@ -55,14 +55,16 @@ void MotionPlanningFrame::planButtonClicked()
 void MotionPlanningFrame::executeButtonClicked()
 {
   ui_->execute_button->setEnabled(false);
-  planning_display_->addBackgroundJob(boost::bind(&MotionPlanningFrame::computeExecuteButtonClicked, this), "execute");
+  // execution is done in a separate thread, to not block other background jobs by blocking for synchronous execution
+  planning_display_->spawnBackgroundJob(boost::bind(&MotionPlanningFrame::computeExecuteButtonClicked, this));
 }
 
 void MotionPlanningFrame::planAndExecuteButtonClicked()
 {
   ui_->plan_and_execute_button->setEnabled(false);
   ui_->execute_button->setEnabled(false);
-  planning_display_->addBackgroundJob(boost::bind(&MotionPlanningFrame::computePlanAndExecuteButtonClicked, this), "plan and execute");
+  // execution is done in a separate thread, to not block other background jobs by blocking for synchronous execution
+  planning_display_->spawnBackgroundJob(boost::bind(&MotionPlanningFrame::computePlanAndExecuteButtonClicked, this));
 }
 
 void MotionPlanningFrame::stopButtonClicked()
--- a/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
+++ b/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
@@ -149,6 +149,9 @@ void MotionPlanningFrame::computePlanAndExecuteButtonClicked()
   if (!move_group_)
     return;
   configureForPlanning();
+  // move_group::move() on the server side, will always start from the current state
+  // to suppress a warning, we pass an empty state (which encodes "start from current state")
+  move_group_->setStartStateToCurrentState();
   ui_->stop_button->setEnabled(true);
   bool success = move_group_->move();
   onFinishedExecution(success);
--- a/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
+++ b/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
@@ -162,63 +162,66 @@ void MotionPlanningFrame::updateQueryStateHelper(robot_state::RobotState &state,
     configureWorkspace();
     if (const robot_model::JointModelGroup *jmg = state.getJointModelGroup(planning_display_->getCurrentPlanningGroup()))
       state.setToRandomPositions(jmg);
+    return;
   }
-  else
-    if (v == "<random valid>")
-    {
-      configureWorkspace();
 
-      if (const robot_model::JointModelGroup *jmg =
+  if (v == "<random valid>")
+  {
+    configureWorkspace();
+
+    if (const robot_model::JointModelGroup *jmg =
         state.getJointModelGroup(planning_display_->getCurrentPlanningGroup()))
+    {
+      // Loop until a collision free state is found
+      static const int MAX_ATTEMPTS = 100;
+      int attempt_count = 0; // prevent loop for going forever
+      while (attempt_count < MAX_ATTEMPTS)
       {
-        // Loop until a collision free state is found
-        static const int MAX_ATTEMPTS = 100;
-        int attempt_count = 0; // prevent loop for going forever
-        while (attempt_count < MAX_ATTEMPTS)
-        {
-          // Generate random state
-          state.setToRandomPositions(jmg);
+        // Generate random state
+        state.setToRandomPositions(jmg);
 
-          state.update(); // prevent dirty transforms
+        state.update(); // prevent dirty transforms
 
-          // Test for collision
-          if (planning_display_->getPlanningSceneRO()->isStateValid(state, "", false))
-            break;
+        // Test for collision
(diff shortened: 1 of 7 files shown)


visualization/planning_scene_rviz_plugin/include/moveit/planning_scene_rviz_plugin/planning_scene_display.h
```diff
@@ -88,7 +88,7 @@ class PlanningSceneDisplay : public rviz::Display
   void queueRenderSceneGeometry();
 
   // pass the execution of this function call to a separate thread that runs in the background
-  void addBackgroundJob(const boost::function<void()> &job, const std::string &name);
+  void addBackgroundJob(const boost::function<void()> &job, const std::string &name, bool ownThread=false);
```
rhaschke: Due to this change, the PR is not ABI compatible. However, this is only related to the rviz plugin, which shouldn't be linked elsewhere. However, if in doubt, we can also target jade-devel.

davetcoleman: The only thing that bothers me about this is that its not using ROS's underscore variable convention ;-)

move_group/src/default_capabilities/execute_service_capability.cpp
```diff
@@ -45,6 +45,7 @@ move_group::MoveGroupExecuteService::MoveGroupExecuteService():
 
 void move_group::MoveGroupExecuteService::initialize()
 {
+  // TODO: Use own callback queue and spinner thread to not block on execution?
```
davetcoleman: +1

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
```diff
@@ -162,63 +200,66 @@ void MotionPlanningFrame::updateQueryStateHelper(robot_state::RobotState &state,
     configureWorkspace();
     if (const robot_model::JointModelGroup *jmg = state.getJointModelGroup(planning_display_->getCurrentPlanningGroup()))
       state.setToRandomPositions(jmg);
+    return;
```
davetcoleman, line 203: So in this function `updateQueryStateHelper()` you don't actually change any logic but rather just wanted to cleanup the program flow?

visualization/planning_scene_rviz_plugin/src/planning_scene_display.cpp
```diff
@@ -227,9 +227,12 @@ void PlanningSceneDisplay::reset()
   }
 }
 
-void PlanningSceneDisplay::addBackgroundJob(const boost::function<void()> &job, const std::string &name)
+void PlanningSceneDisplay::addBackgroundJob(const boost::function<void()> &job, const std::string &name, bool ownThread)
```
davetcoleman: again please use underscores

davetcoleman:
I think its a good idea to address https://github.com/ros-planning/moveit_ros/pull/709#discussion_r70888457 but I understand if you don't have the cycles for it right now

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
```diff
@@ -55,14 +55,20 @@ void MotionPlanningFrame::planButtonClicked()
 void MotionPlanningFrame::executeButtonClicked()
 {
   ui_->execute_button->setEnabled(false);
-  planning_display_->addBackgroundJob(boost::bind(&MotionPlanningFrame::computeExecuteButtonClicked, this), "execute");
+  boost::thread(boost::bind(&MotionPlanningFrame::computeExecuteButtonClicked, this));
```
davetcoleman: Why use boost threads when this class already has its own background job framework setup? Seems like a departure of the functional style of the rest of the class.

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
rhaschke, line 203: Yep. I found it difficult to follow the nested if structure.

visualization/planning_scene_rviz_plugin/include/moveit/planning_scene_rviz_plugin/planning_scene_display.h
rhaschke: Where do you want to have the underscores? As far as I understand it, only member variables should be suffixed by an underscore. Is there anything else?

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
rhaschke: Executing the motion in the (single) background thread blocks visual updates of the scene robot.

Unfortunately, this is only mentioned in the commit log. I will add a source-code comment.

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
```diff
@@ -55,14 +55,22 @@ void MotionPlanningFrame::planButtonClicked()
 void MotionPlanningFrame::executeButtonClicked()
 {
   ui_->execute_button->setEnabled(false);
-  planning_display_->addBackgroundJob(boost::bind(&MotionPlanningFrame::computeExecuteButtonClicked, this), "execute");
+  // executing the motion in the (single) background thread, blocks visual updates of the scene robot
+  boost::thread(boost::bind(&MotionPlanningFrame::computeExecuteButtonClicked, this));
```
v4hn: Why are visual updates blocked? I suppose planning scene updates don't get through?
Maybe we could have two callback queues, one for callback related functions and one for handling of user interaction?
I suppose it is a good idea to have a uniform way to run background computation on button pressed throughout rviz-moveit.

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
```diff
@@ -129,16 +137,48 @@ void MotionPlanningFrame::computePlanButtonClicked()
 void MotionPlanningFrame::computeExecuteButtonClicked()
 {
   if (move_group_ && current_plan_)
-    move_group_->execute(*current_plan_);
+  {
+    // ui_->stop_button->setEnabled(true); // enable stopping
```
v4hn: Why is this commented?

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
```diff
@@ -129,16 +137,48 @@ void MotionPlanningFrame::computePlanButtonClicked()
...
+    bool success = move_group_->execute(*current_plan_);
+    ui_->result_label->setText(success ? "Executed" : "Execution failed");
+    ui_->stop_button->setEnabled(false); // disable stopping
+  }
+
+  // Hm. The current state is only updated after a while (only when using /execute/ service)
+  // TODO: Wait for update to planning scene? Related to #671
+  sleep(1);
+  updateStartStateToCurrent();
```
v4hn: If this does not explicitly ask move_group for the current state via a service request, :-1:.
Waiting for the current state via `sleep` is a workaround that might fail on slow systems/heavy load.
This is the exact problem I was trying to address by only sending empty start states in #710 .

In my opinion, a proper solution  would add a new `<always-current>` option and keep updating the maintained start state in this case. This has some non-trivial edge-cases, but it would allow for other nodes to move the arm without rendering the display's start-state obsolete.

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
```diff
@@ -129,16 +137,48 @@ void MotionPlanningFrame::computePlanButtonClicked()
...
     return;
   configureForPlanning();
-  move_group_->move();
+  ui_->stop_button->setEnabled(true);
+  bool success = move_group_->move();
+  if (success) ui_->result_label->setText("Executed");
+  else ui_->result_label->setText(!ui_->stop_button->isEnabled() ? "Stopped" : "Failed");
   ui_->plan_and_execute_button->setEnabled(true);
+  ui_->stop_button->setEnabled(false);
+  updateStartStateToCurrent();
```
v4hn: Here, you're not sleeping and the start state might be obsolete, right?

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
v4hn, line 203: +1, the nested structures throughout MoveIt! are totally confusing and I would like to see most of them replaced by "Check and early return", as you did here.

visualization/planning_scene_rviz_plugin/include/moveit/planning_scene_rviz_plugin/planning_scene_display.h
v4hn: all other argument names in this file use underscore notation, so please change `ownThread` to `own_thread`.

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
davetcoleman, line 203: +1 I'm a big fan of check and early return

visualization/planning_scene_rviz_plugin/include/moveit/planning_scene_rviz_plugin/planning_scene_display.h
v4hn: I agree that breaking ABI in this code is well-justified by safety concerns.
You are not using `addBackgroundJob` anywhere though. :-)
Also, it might be a good idea to have a separate function name for that, instead of extending the existing one with orthogonal code. `spawnBackgroundJob`?
Alternatively, we could simply stick with the `boost::thread` one-liner you used in this request.
But then it might still be a good idea to have explicit comments above each of these lines to explain why this should not be `addBackgroundJob`.

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
rhaschke: Because it's currently not working. A service cannot be interrupted. @wkentaro suggested an option which I will try later.

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
rhaschke: I know, that this is a work-around for now. This should addressed with #716.
I can only address one thing at a time ;-)

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
rhaschke: Using the move_group action server, the issue wasn't that dramatic, because intermediate updates come through.
However, you are right, in principle the delayed update issue is observable here too. Should be fixed in #716 - hope a lot of people will help on this ;-)

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
rhaschke, line 203: Nice, that we agree on that style ;-)

visualization/planning_scene_rviz_plugin/include/moveit/planning_scene_rviz_plugin/planning_scene_display.h
rhaschke: @v4hn @davetcoleman Whatever you prefer. Please vote ;-)

visualization/planning_scene_rviz_plugin/src/planning_scene_display.cpp
rhaschke: Got it.

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
rhaschke: Without separate threads for this, the background-job-thread was mainly blocked with waiting for the trajectory execution to finish. Everything else going to the background job simply needs to wait as well - which doesn't make sense.
To have a more uniform "look&feel", I augmented `addBackgroundJob()` with an extra option `ownThread` defaulting to false. For some reason the use of this code got lost when cleaning up my messy local repo. However, you suggested to go with boost::thread only. I have a slight preference for that too. Please vote for an option!

visualization/planning_scene_rviz_plugin/include/moveit/planning_scene_rviz_plugin/planning_scene_display.h
v4hn: +1 on a one-line `spawnBackgroundJob` function and a comment above that explains why this sometimes makes sense to use instead of `addBackgroundJob`.
I would be interested in a real explanation myself :-)

davetcoleman: +1 to @v4hn's comment

rhaschke:
I addressed all comments. Remaining issue is to turn ExecuteService in ExecuteAction and of course deal with #716.

rhaschke:
Further cleaned up and rebased.
Disabling QueryStartState by default as suggested in #710.
Relies on #717.
For correct working of start-state updates, we need to fix #716 as well.
Could be merged now.

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
```diff
@@ -129,18 +137,46 @@ void MotionPlanningFrame::computePlanButtonClicked()
...
+void MotionPlanningFrame::computeStopButtonClicked()
+{
+  if (move_group_)
+    move_group_->stop();
+}
+
+void MotionPlanningFrame::onFinishedExecution(bool success)
+{
+  // visualize result of execution
+  if (success) ui_->result_label->setText("Executed");
```
davetcoleman: you should line break after if() statement per ROS style

davetcoleman:
+1

visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
rhaschke: Fixed.

davetcoleman:
since this is a relatively major change (new GUI button) to a stable (indigo) release I think you should announce this on the mailing list

also, please be sure to cherry-pick to jade and kinetic

thanks!

wkentaro:
@davetcoleman @rhaschke Thanks for working for this!

davetcoleman:
the thanks goes to @rhaschke !

rhaschke:
> please be sure to cherry-pick to jade and kinetic

Dave, you already merge-committed #718 into jade-devel. I would have loved to use a squashing commit only, i.e. forwarding jade-devel to the cherry-pick branch.
Anybody minds to rebase jade-devel there?

The same I would do for kinetic. For me this eases reading the git graph: 
Merge commits are unique contributions, while cherry-picks could be committed directly. 
I just had to use a PR to let Travis do the compile checking.

rhaschke:
> since this is a relatively major change (new GUI button) to a stable 
> (indigo) release I think you should announce this on the mailing list

Dear Dave,
before announcing on the mailing list, I would like to finish #716, 
which is a major pre-requisite for #713 to work correctly.
Hopefully, I find some time this evening to continue on that.
In the same vein, I'm planning to merge those changes to Kinetic only as 
a bundle.

davetcoleman:
> I just had to use a PR to let Travis do the compile checking.

Please make note of that in your PR if you don't want it actually merged yet.

> Dave, you already merge-committed #718 into jade-devel.

Yes, I hadn't seen the PR until after I made that comment

You're announcement timeline sounds good, thanks!

PR #727, opened 2016-07-26
Changing minimum value for jump_factor to 0.0

Jntzko:
We should be able to set the minimum value for jump_factor to 0 to disable it.
So this change in the config allows us to disable this feature.

In the robot_state.cpp in line 1740 is a test if the value is greater then 0.0, so there should be the possibility to set this value to 0.

`bool test_joint_space_jump = jump_threshold > 0.0;`


v4hn:
Thanks for the request!
This seems to be a simple typo in the config.

I'll also cherry-pick this to jade and kinetic.

